PR #29, opened 2016-11-28
Refined release.sh

butomo1989:


butomo1989:
closed. reason: it do not support to create PR from forked repo yet

Issue #40, opened 2016-12-20
ContainerNotFoundException when setting up nodes


arielschiavoni:
Hi there, first of all thanks for the work done on this very helpful tool.

I tried `zalenium` a couple of days ago and everything worked out of the box. However today when I run zalenium I get `ContainerNotFoundException` which seems to prevent the containers from been created an attached to the grid/hub.

```
docker run --rm -ti \
    --name selenium \
    -p 4444:4444 \
    -p 5555:5555 \
    -e DOCKER=1.12 \
    -v /var/run/docker.sock:/var/run/docker.sock \
    -v /tmp/videos:/home/seluser/videos \
    dosel/zalenium:2.53.1k start
```

I noticed `zalenium:2.53.1k` has been recently released so I also tried to run `zalenium:2.53.1j` which throws the same exception.

```
11:51:25.558 INFO - [DS]  Setting up 2 nodes...
11:51:27.418 ERROR - [DS] com.spotify.docker.client.exceptions.ContainerNotFoundException: Container not found: da227fb1fe4f543c124a59f8e79237538116e22032597f49f5fd10e8f4237043
com.spotify.docker.client.exceptions.ContainerNotFoundException: Container not found: da227fb1fe4f543c124a59f8e79237538116e22032597f49f5fd10e8f4237043
	at com.spotify.docker.client.DefaultDockerClient.containerAction(DefaultDockerClient.java:613)
	at com.spotify.docker.client.DefaultDockerClient.startContainer(DefaultDockerClient.java:601)
	at de.zalando.tip.zalenium.proxy.DockerSeleniumStarterRemoteProxy.startDockerSeleniumContainer(DockerSeleniumStarterRemoteProxy.java:395)
	at de.zalando.tip.zalenium.proxy.DockerSeleniumStarterRemoteProxy$1.run(DockerSeleniumStarterRemoteProxy.java:436)
Caused by: com.spotify.docker.client.exceptions.DockerRequestException: Request error: POST unix://localhost:80/containers/da227fb1fe4f543c124a59f8e79237538116e22032597f49f5fd10e8f4237043/start: 404
	at com.spotify.docker.client.DefaultDockerClient.propagate(DefaultDockerClient.java:2084)
	at com.spotify.docker.client.DefaultDockerClient.request(DefaultDockerClient.java:2045)
	at com.spotify.docker.client.DefaultDockerClient.containerAction(DefaultDockerClient.java:609)
	... 3 more
```

**Environment**
[...]

PR #41, opened 2016-12-20
Make it clear in README.md 'Running it' section that container must be called zalenium

arielschiavoni:
Closes [40](https://github.com/zalando/zalenium/issues/40)


diemol:
Thanks! Merging right away since it is a modification in the README. The build will fail because we need to improve something with the environment variables in Travis.

arielschiavoni:
Thanks!

PR #46, opened 2017-01-02
Add support to use TestingBot.com Selenium grid

testingbot:
This adds support for TestingBot.com which is a large online Selenium grid, similar to BrowserStack and SauceLabs.

Please let me know if this needs any changes and I'll modify the pull request.

Thanks!


diemol:
Hi @testingbot,

Thanks for this PR! This level of interest in our project makes us very happy :)

We can gladly accept the PR, but can you please help us first with:

- Checking the error in the build? It seems to be a compilation error. A parameter is missing for a method.
- Is it possible to get an open source account from TestingBot? We would need it to run the integration tests in Travis and to check that the routing is being done properly.

Please let me know if you need any support for these two things.

Cheers,

diemol:
Hi again,

I found the link to create a open source account with TestingBot, with my gmail user "diemol". I have now a "free-trial", but I cannot find where to submit a request to change the account to "open source".

Could you please help me with that?

diemol:
I changed the target branch so we can run the integration tests after merging (once I get the user and key for TestingBot), if everything goes well there, then we can merge to master and Selenium 3 branches.

diemol:
Hi @testingbot

After merging and making a few changes, most things are working. Nevertheless there is a problem with the tunnel for local testing.

The tunnel starts without problems and I can see that it has been established with our checks and also through the TestingBot website.

After reading [this](https://testingbot.com/support/other/tunnel), I notice that it is needed to use http://localhost:4445 instead of http://hub.testingbot.com:80. Note that when the TestingBotProxy is registered, the second url is used for `getRemoteHost`.

The problem comes because I have not found a way to redirect the incoming request to http://localhost:4445 when the tunnel is on. One way SauceLabs and BrowserStack do it, is by allowing the user to put a capability in the test pointing to the tunnel identifier.

Do you have any ideas to make the tunnel work properly? Any suggestion would be appreciated.

We'll try to think here to see if we come up with something.

Another option would be to integrate TestingBot without the tunnel option for now, and think about it later.

diemol:
Thanks for the hint, it is working ☺️

diemol:
Check https://github.com/zalando/zalenium/pull/50 for progress.

PR #60, opened 2017-01-31
Increase max body size nginx

fffilimonov:
Tests for file upload fails.
Default value for nginx is 1M.
So when upload more you get 413 error.
300M shuold be enough for everyone ;)

But further it's better to make configurable


diemol:
Hi @fffilimonov,

Thanks for this PR! One more use case that we did not consider...

We will merge it soon to a new branch so we can run the integration tests and then merge it to master.

Issue #73, opened 2017-03-09
Orphan containers are created after a cleanup loop


yaronassa:
I use the grid to work with a firefox and chrome containers, and end my sessions between tests.
In the 1st test, the initial containers are used, then disposed, and everything's great.
Then I start another test cycle, and a redundant chrome container is created, though no session is opened on it.
This behavior is reproduced for every test cycle, until I reach the max container limit, and the new session request is queued forever.

The log with my comments in [[[]]]:
```
10:19:37.395 INFO - [DS] Capabilities grabbed from the docker-selenium image
10:19:37.401 INFO - The node is registered to the hub and ready to use
10:19:37.407 INFO - Registered a node http://172.17.0.8:30000
10:19:37.417 INFO - [DS]  Setting up 2 nodes...
10:19:38.122 INFO - 2 containers were created, it will take a bit more until all get registered.
10:19:43.391 INFO - Registered a node http://localhost:40000
10:19:43.869 INFO - Registered a node http://localhost:40001
10:20:00.511 INFO - Got a request to create a new session: Capabilities [{rotatable=true, locationContextEnabled=true, loggingPrefs=org.openqa.selenium.logging.LoggingPreferences@212e1d7f, browserName=firefox, javascriptEnabled=true, handlesAlerts=true, requestOrigins={name=webdriverio, version=4.6.1, url=http://webdriver.io}}]
10:20:00.516 INFO - Trying to create a new session on test slot {firefox_binary=/home/seluser/firefox-for-sel-3/firefox, seleniumProtocol=WebDriver, acceptSslCerts=true, browserName=firefox, maxInstances=1, version=51.0.1, platform=LINUX}
10:20:00.531 INFO - Got a request to create a new session: Capabilities [{rotatable=true, locationContextEnabled=true, loggingPrefs=org.openqa.selenium.logging.LoggingPreferences@3f126dc1, browserName=chrome, javascriptEnabled=true, handlesAlerts=true, requestOrigins={name=webdriverio, version=4.6.1, url=http://webdriver.io}}]
10:20:00.550 INFO - localhost:40001 start-video: Video recording is disabled
10:20:00.569 INFO - Trying to create a new session on test slot {chrome_binary=/usr/bin/google-chrome-stable, seleniumProtocol=WebDriver, acceptSslCerts=true, browserName=chrome, maxInstances=1, version=56.0.2924.87, platform=LINUX}
10:20:00.570 INFO - localhost:40000 start-video: Video recording is disabled
[...]

PR #80, opened 2017-03-16
Converting FLV videos to MP4 (when downloading from SauceLabs)

diemol:
... also improving how containers get created when the test requests arrive, avoiding leaving many orphaned containers #73 (not final solution but it is really close to it).


diemol:
👍

scherniavsky:
:+1:

PR #84, opened 2017-03-21
Adding support for custom screen resolution in the capabilities

diemol:
Also
* Upgrading Sauce Connect to version 4.4.5
* Improving how containers get created to avoid having too many orphan containers


elgalu:
:+1:

diemol:
👍

PR #113, opened 2017-05-12
Fixed grep for container id when running in DinD

mattcfox:
The way that the container id was previous found was broken when using zalenium with Docker in Docker, and would retrieve the DinD container's id.

### DIFF
--- a/scripts/zalenium.sh
+++ b/scripts/zalenium.sh
@@ -256,7 +256,7 @@ trap DockerTerminate SIGTERM SIGINT SIGKILL
 StartUp()
 {
     EnsureDockerWorks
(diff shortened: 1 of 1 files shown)


diemol:
Hi @mattcfox, thanks for this PR.

Just want to double check if I understood, you are running Zalenium inside a docker container, and all the other docker-selenium containers are created inside that initial container?

elgalu:
:+1:

diemol:
👍

elgalu:
:+1:

diemol:
Thanks for the PR @mattcfox, we tested it and it works well :)

mattcfox:
@diemol, apologies for the late reply.  Yes, I'm running zalenium as well as all of the containers it spawns within a single container.  I'm using `docker:stable-dind`.

PR #114, opened 2017-05-14
Html templates

diemol:
Rendering custom proxies and console servlets with html templates instead of mixing Java and HTML code.


elgalu:
:+1:

diemol:
👍

PR #128, opened 2017-05-27
Adding factory to create a "container client"...

diemol:
... which serves as an abstraction to the current docker client, this is the initial step in order to support Kubernetes.


diemol:
👍

elgalu:
👍

Issue #103, opened 2017-04-22
kubernetes support


pearj:
Hi,

I'm wondering if there are any plans to add kubernetes support?

I realise that zalenium as it stands is very tied to using the docker command directly.  But with a bit of refactoring, I think it wouldn't be too hard to have a zalenium-docker and a zalenium-kubernetes version if the core bits of zalenium were moved to a separate maven artifact and the docker specific parts moved to another maven artefact, which would allow a kubernetes version.

Thoughts?

elgalu:
Hi, as of now we are using https://github.com/spotify/docker-client to interact with Docker.

To work on a kubernetes version I believe we would need a kubernetes-java client, as of now there only seem to exist go and python clients:
https://github.com/kubernetes/client-go
https://github.com/kubernetes-incubator/client-python

elgalu:
Ah, found non-official java clients:

https://github.com/fabric8io/fabric8/tree/master/components/kubernetes-api
https://github.com/fabric8io/kubernetes-client

pearj:
Hi,

I had the fabric8 kubernetes-client in mind as well when I created this issue.
The thing that is cool about it, is that it has support for authenticating via a service account which means that the authentication to the Kubernetes API should be transparent, as long as the zalenium pod is created with a service account that has enough permissions to be able to create and delete pods, and run commands against a pod.

Starting a selenium Pod would probably look something like this:
```java
DefaultKubernetesClient client = new DefaultKubernetesClient();
Pod newPod = client.pods()
    .inNamespace("rio")
    .createNew()
    .withNewMetadata()
        .withGenerateName("selenium-node")
        .addToLabels("app", "zalenium")
    .endMetadata()
    .withNewSpec()
        .addNewContainer()
            .withName("selenium-node")
            .withImage("elgalu/selenium")
            .addNewEnv()
                .withName("Bob")
                .withValue("The builder")
            .endEnv()
        .endContainer()
        .withRestartPolicy("Never")
    .endSpec()
    .done();
```

Does Zalendo use Kubernetes at the moment? Or do you not have any Kubernetes capability at the moment?

diemol:
Zalando is moving to Kubernetes, so we will start checking into this soon. Nevertheless any help is highly appreciated :)

dspasojevic:
@diemol if you are able to provide some rough direction for how you would like docker and kubernetes support to co-exist @pearj and I may be able to action this.

diemol:
[...]

PR #138, opened 2017-06-06
First cut of kubernetes support #103

pearj:
What works now:

* Creating containers
* Deleting containers
* Live preview
* Exec commands
* Copying files
* Mounting the shared volume

What is missing:
* Documentation


diemol:
Hi @pearj,

Thanks for the PR!

We will merge it after we release the changes we are doing for performance, hopefully end of this week.

On the mean time, is it possible for you to add a markdown file with some setup instructions?

pearj:
Sure, I'll add some doco. 

Hopefully by the end of the week I'll have solved file copying. Fingers crossed tomorrow.

pearj:
I got copying files working. Turns out I had the tar command syntax wrong, whoops!  Also I ran into bugs with kubernetes-client how it handles streams being shutdown in failure scenarios.

But I have it working now, yay.

Tomorrow I think I'll add a flag "--minSpareCapacity" or something.  Because the time to start a node in kubernetes is pretty slow, so kubernetes will need something like that.

I'll write some doco tomorrow too

pearj:
I've now added support for auto-mounting the shared drive across all the nodes.

I've started on the Doco, but because we use OpenShift here I'm using some features to start zalenium that are OpenShift specific, so I'll have to find out what works in Kubernetes.

The kubernetes support inside Zalenium itself is only using the kubernetes api at the moment, so that's safe. It's just starting the zalenium server itself, I'll have to find find some examples that work in native kubernetes.

MiluchOK:
If you tell me how to use it, I'll sell you my soul and all my belongings!

diemol:
Hi @pearj, we just released 3.3.1i, which is the work we did during the last two weeks. 
I'll review this PR quickly tomorrow and I'll merge it to the new branch I created, so after that we can start trying it here. I guess that Minikube will do for the initial testing.

Again, thanks a lot for the effort you have put on this!

pearj:
Lol @MiluchOK.  I promise I'll write some doco today.

pearj:
Sooo, I don't think doco will come till next week :(.  My workstation, which is a virtual machine, has died and being Friday afternoon here all the people that can fix have gone for the long weekend.  So it will have to wait until Tuesday.

I have fixed a few other bugs, but I can't commit them because they're still on that workstation.

I also made it possible to override the selenium image, which is normally "elgalu/selenium".  I found that it's pretty slow if it has to check with docker hub ever time a node starts, especially from Australia at least. It adds about 5 extra seconds to the start time when there is no update.  I found that specifying a private registry is a lot faster, plus it means you can control when your selenium nodes should update.

Since I use OpenShift, I did a little bit of research about what is available with plain kubectl, and it looks like you can get most of it done with something like `kubectl run zalenium-test --image=172.23.192.79:5000/delivery/zalenium --env="http_proxy=" --env="https_proxy=" --env="HTTP_PROXY=" --env="HTTPS_PROXY=" --overrides='{"spec": {"template": {"spec": {"serviceAccount": "zalenium"}}}}' -l app=zalenium-test --port=4444 -- start --firefoxContainers 0 --chromeContainers 2`

That only exposes port 4444 though, so either it will need a yaml file or an kubectl patch, to add port 5555 as well.

diemol:
👍

elgalu:
👍

pearj:
Thanks!

diemol:
Thanks for your effort @pearj!
We'll play with during this week, and help with all the TODOs.

diemol:
I pushed one commit to the kubernetes brach to fix a little thing with the containerId method, which didn't let the the whole video recording etc... work. 

The first thing I was doing was just to check that Zalenium in docker was still working, and it looks that it works well 👍 

Now we can actually focus on checking it in kubernetes during this week, looking forward to that!

Issue #148, opened 2017-06-09
Chrome Cert Support, Add Ability to Mount Arbitrary Folders in Node File System


ReservedDeveloper:
This closely mimics #64, but I'd like to expand on it. More specifically, I'm looking to configure Chrome's cert store at `~/.pki/nssdb`.

Being able to mount a volume via, say, `-v /my-host-dir:/tmp/node/node-file-path`, might be the most straightforward. Though it is handing library users a loaded gun, so to speak.

Expanding the mount behavior to a more general scope seems like the most versatile solution, but I'd also be thrilled just to see `$HOME/.pki/nssdb/` modification ability.

ReservedDeveloper:
For context, [this link](https://chromium.googlesource.com/chromium/src/+/lkcr/docs/linux_cert_management.md) details the cert process for Chrome a little more in depth.

pearj:
Instead of editing the chrome certs db, why don't you just set the "[acceptSslCerts](https://github.com/SeleniumHQ/selenium/wiki/DesiredCapabilities)" capability to true?

ReservedDeveloper:
@pearj A valid question. I'm actually interested in utilizing client/personal certs that are being presented _to_ the server for authentication purposes. The setting you mention is for handling CA certs presented _from_ the server. Because of the way it's entangled in OS-level security (esp. on Windows), it's not something Selenium presents an easy way to handle.

There _is_ a setting for presenting certificates based on certain criteria ([AutoSelectCertificateForUrls](https://www.chromium.org/administrators/policy-list-3#AutoSelectCertificateForUrls)), but it still requires a cert be loaded in the store.

pearj:
Ahh mutual ssl. Fair enough

pearj:
I was thinking on kubernetes this could be fairly easily supported by mounting any persistent volume that is mounted on zalenium that has an access mode of ReadOnlyMany or ReadWriteMany. 

See:
https://kubernetes.io/docs/concepts/storage/persistent-volumes/#access-modes

But you're not using kubernetes are you?

ReservedDeveloper:
No, I'm using Docker. But I'll keep that in mind for the PR I'm putting together.

diemol:
Released https://github.com/zalando/zalenium/releases/tag/3.3.1l

Thanks for contributing @ReservedDeveloper!

PR #154, opened 2017-06-18
Feature #148 Add add't mount points in node

ReservedDeveloper:
* Added mount points from `/tmp/node/[TARGET_FOLDER_ON_NODE]`
  * Updated documentation

### DIFF
--- a/docs/usage_examples.md
+++ b/docs/usage_examples.md
@@ -115,21 +115,27 @@ explicitly tell Zalenium which major version you are using via `-e DOCKER=1.11` 
       --privileged dosel/zalenium start --screenWidth 1440 --screenHeight 810 --timeZone "America/Montreal"
   ```
 
-### with a multi-purpose folder mounted
-This is a folder that you can mount as a volume when starting Zalenium, and it will be mapped across all the docker-selenium containers. 
-It could be used to provide files needed to run your tests, such as filed that need to be open from the browser or folders to use when 
-starting Chrome with a specific profile.
+### with node folders mounted
+
+This is a collection of folders that you can mount as volumes when starting Zalenium by prefixing the destination with `/tmp/node/`, and it will be mapped across 
+all the docker-selenium containers from the root folder after stripping the `/tmp/node` prefix. 
+So `-v /your/local/folder:/tmp/node/home/seluser/folder` will map to `/home/seluser/folder` on the node. 
+It can be used to provide further customization to your nodes, such as adding client certificates 
+for your browser, or mimicking prior multi-purpose folder, both shown below.
 
   ```sh
     docker run --rm -ti --name zalenium -p 4444:4444 -p 5555:5555 \
       -v /var/run/docker.sock:/var/run/docker.sock \
       -v /tmp/videos:/home/seluser/videos \
-      -v /your/local/folder:/tmp/mounted \      
+      -v /your/local/folder/with/certStore:/tmp/node/home/seluser/.pki/nssdb \      
+      -v /your/local/folderB:/tmp/node/home/seluser/folderB \      
+      -v /tmp/mounted:/tmp/node/tmp/mounted \
       --privileged dosel/zalenium start 
   ```
-After starting Zalenium with this mounted volume, any file created in the host in `/your/local/folder`, will be available in
-`/tmp/mounted` across all containers. Please note that the folder name in the host can be any you want, the important part is 
-to map properly.
+
+Please take caution in mounting system folders such as `/etc`, as this behavior has not been tested with such configuration.
+
+**NOTE:** There are certain protected points which cannot be mounted via `/tmp/node`. See `DockerContainerClient#PROTECTED_NODE_MOUNT_POINTS`.
 
 ### More configuration parameters
 
--- a/src/main/java/de/zalando/ep/zalenium/container/ContainerClient.java
+++ b/src/main/java/de/zalando/ep/zalenium/container/ContainerClient.java
@@ -6,7 +6,7 @@ import java.util.Map;
 
 public interface ContainerClient {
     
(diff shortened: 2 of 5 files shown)


ReservedDeveloper:
I was not able to run the integration tests locally, but was able to test against a [demo project](https://github.com/geb/geb-example-maven) without issue and see the mount points in question.

That said, this does have a slightly broad approach, so totally understand if you'd rather this be a little narrower in scope (say, perhaps, only able to mount folders within `/home/seluser`)

diemol:
Hi @ReservedDeveloper,

Thanks for this PR!

If I understand correctly, this addresses #148, and it is done like that because the Chrome certs need to be on that specific folder?

If so, perhaps we should take the approach to map all the mount points provided in the `docker run` except the docker socket and the `/home/seluser/videos` folder. Like that the users could mount any folders they want on any location. I am just afraid that we start adding special cases one by one, and it is maybe better to add them all in a flexible way.

Having said that, could you please check the PR again after we release the 1st beta for Kubernetes? (Should happen tomorrow). I expect some merge conflicts after the release for this PR. After that I would be happy to merge it.

ReservedDeveloper:
@diemol That is correct. And I think what you're proposing (minus the protections for `/videos` and the socket) is how it's currently set up. The way it's currently oriented, the node mount simply has the `/tmp/node` stripped from its bind point. So:

* `docker run -v /my/folder:/tmp/node/home/seluser/foobar` will map to each node's `/home/seluser/foobar` folder.
* `docker run -v /my/folder:/tmp/node/tmp/mounted` would mimic the current multi-purpose volume share.
* `docker run -v /my/folder:/tmp/node/etc` would replace the `/etc` folder with the host folder, and perhaps make for some unknown complications, which I attempted to call out in the documentation update, but could be reworded.

ReservedDeveloper:
@diemol Got the latest from master merged in locally. As I'm adding in protections, how would you like mounting a protected point (ex via: `docker run -v /my/local/folder:/tmp/node/home/seluser/videos`) to be handled? 

It could be more permissive, and simply not trim the `/tmp/node` piece. We'd have some inconsistent node mount behavior, but as it's not necessarily a hindrance to operation to mount in the `/tmp` directory, this may be preferred.

Alternatively, could just throw an `IllegalArgumentException` if one of those directories is encountered. Thoughts?

diemol:
To be honest, when I checked the PR the first time I didn't catch the `/tmp/node` part, sorry about that. 

But the only question I have is: why do we need to have that `/tmp/node` prefix? Can't we just leave them as it is since we will filter a couple of ones?

Besides from `/home/seluser/videos` and the docker socket, I would say that we wound't need to filter anything else.

ReservedDeveloper:
I suppose I was thinking of the feature more as 'opt in' than 'opt out'. Help prevent any unexpected issues due to mounting. Or if you were wanting to mount _only_ in the zalenium container, but not the nodes.

If we go that route, I'd likely just skip any conflicting mount points, and add a log stating why it was skipped.

diemol:
@ReservedDeveloper, I think you made a good point. Perhaps having `/tmp/node` as a prefix is a good idea to avoid mounting issues, as you mention. We also "protect" the main Zalenium container in that way I guess.

I think we could go that way, and just document it properly in the current document and in the release (since this is a breaking change for existing users of this feature).

Could you please try to add a unit test for it? I am not sure how easy it is, perhaps we need to refactor something.

ReservedDeveloper:
I'll add that in with the protection updates. I should have the time to put that together tomorrow evening. 

@pearj mentioned an approach for adding the mapping feature to the Kubernetes side that I'm hoping to add in as well, though we could always split that out to a separate ticket.

pearj:
@ReservedDeveloper You shoudl probably keep the kubernetes support for arbitrary mounts a separate ticket I reckon. Especially if you aren't able to test it, and we don't have travis running automatic integration tests for kubernetes yet.

diemol:
Sounds good @ReservedDeveloper, thanks!

 Thanks @pearj for the extra pair of eyes :)

ReservedDeveloper:
Updated to resolve the latest from master, and to add the discussed protections. However, in looking at `DockerContainerClient#createContainer`, building up a unit test may be a little tricky, as you suggested. In particular, the `HostConfig.builder()` and `ContainerConfig.builder()` calls put a wrench in things. I'd considered extracting those out to a wrapper class or something, but that sort of refactoring felt a little out of scope.

ReservedDeveloper:
@diemol not sure how you'd like to proceed on testing. That said, I'd like to take some time this weekend to rebase and clean up the history a touch (as well as resolve my listed email address), so please hold on merging this if you wouldn't mind. Just let me know as far as testing goes!

diemol:
@ReservedDeveloper sounds good. I was going to go over this during the weekend, but sure, when you are ready we can check it again and merge it :)

About the tests, just try but if it ends up to be complicated, then just leave it as it is. I think that we just need to refactor that part of the code to make it more testable.

ReservedDeveloper:
@diemol all good to go. Wasn't able to tackle the unit test stuff, but if you'd like to create a new issue I'd be happy to try and put something together in the coming weeks.

diemol:
Hi @ReservedDeveloper, thanks for putting it all together.

I'll merge it to a branch I created to fix an issue with docker-compose and we'll release that in the next couple of days.

About the unit tests, I need to refactor that code to make it more testable. If you can, keep an eye on it so if you feel like, you can add some unit tests in the near future.

One last thing, I still see the issue with the commits where your user is not really visible, you can fix it whenever you want if you want to be seen in the contributors part :)

diemol:
👍

scherniavsky:
:+1:

PR #160, opened 2017-06-18
Release 3.3.1k

### LINKED ISSUE #143: Too many containers are created
After the 3.3.1i release, where the container creation changed to be more aggressive, there are situations where Zalenium creates more containers that the needed ones. This won't affect the tests but it could affect the machine since it is taking resources from it.

The idea is to find a point in between the previous more conservative strategy and the new one.

diemol:
Closes #143 and #103


diemol:
👍

elgalu:
:+1:

PR #174, opened 2017-07-04
Fixing issue 172

diemol:
and also removing static variables for screen size and time zone in order to avoid race conditions...


diemol:
👍

butomo1989:
👍

PR #179, opened 2017-07-09
Allowing --net=host in Linux 

diemol:
and including `mac.host.local` to the hosts file to allow access to the host in OSX. In theory `docker.for.mac.localhost` should have worked but somehow it is not working for Chrome.


diemol:
👍

a1exmac:
👍

PR #183, opened 2017-07-12
Add more detail to the startTunnel option

JMoravec:
In the docs, it is not obvious that that when using the tunnel feature of the cloud testing providers, Zalenium starts the tunnel with the local identifier `zalenium`. (At least for BrowserStack) if this local identifier is used, than the capabilities must have that in the test request to use the tunnel. This updated description (hopefully) clears the usage up.


diemol:
Nice! Thanks @JMoravec!

diemol:
👍

scherniavsky:
:+1:

PR #218, opened 2017-08-30
Creating service for OpenShift

TomLarrow:
Needed to add role=hub to the OpenShift directions, and give a service which specifically selects pods of this role.

I struggled with the directions as written, because the OpenShift directions did not specify a role  the same way the Kubernetes directions did. This was causing the service to bind not only to the zalenium pod, but also to the selenium pods once they spun up.  This caused OpenShift networking to inappropriately direct traffic that was supposed to go to the hub to the nodes.

I also included a zalenium-service.yml which would create the service which properly selects only the zalenium pod.


diemol:
Thanks for this PR @TomLarrow! Just got back from vacation, will review and merge soon.

diemol:
👍

elgalu:
:+1:

Issue #211, opened 2017-08-17
Zalenium spins up lots of containers


mojsha:
I have tried using Zalenium with Minishift and noticed that it seems to spin up alot of containers. 

I then get a "Gateway timeout" error from my client.

pearj:
Can you attached the zalenium container logs?

pearj:
Are you running openshift 3.6 (as minishift)? I haven't tried that version yet. Although when I was writing the initial k8s doco I did try minikube which had k8s 1.6 I thought. So I think zalenium runs ok on 1.6

pearj:
I just started zalenium in minishift 3.6.0 and it started up ok, so that's a bit weird what you're seeing

mojsha:
Ok, I have now downloaded the latest Minishift client and then run it and get this (as was the case initially):


*************************************** Data Processing Agreement ***************************************
--
  | 09:06:45.455 DEBUG - Handling: GET /status
  | 09:06:45.654 DEBUG - Executing: org.openqa.selenium.remote.server.handler.Status@78c3a438)
  | 09:06:45.710 DEBUG - Done: org.openqa.selenium.remote.server.handler.Status@78c3a438
  | 09:06:45.711 DEBUG - Finished: GET /status
  | 09:06:57.449 DEBUG - Handling: GET /status
  | 09:06:57.449 DEBUG - Executing: org.openqa.selenium.remote.server.handler.Status@54d43e5f)
  | 09:06:57.449 DEBUG - Done: org.openqa.selenium.remote.server.handler.Status@54d43e5f
  | 09:06:57.450 DEBUG - Finished: GET /status
  | 09:07:17.282 INFO - Got a request to create a new session: Capabilities [{acceptInsecureCerts=true, browserName=firefox, version=, platform=ANY}]
  | 09:07:17.283 DEBUG - Validating {acceptInsecureCerts=true, browserName=firefox, version=, platform=ANY} in node with capabilities {browserName=firefox, maxInstances=1, platform=LINUX}
  | 09:07:17.289 DEBUG - Available nodes: [http://172.17.0.7:30000]
  | 09:07:17.289 DEBUG - Validating {acceptInsecureCerts=true, browserName=firefox, version=, platform=ANY} in node with capabilities {browserName=firefox, maxInstances=1, platform=LINUX}
[...]

PR #242, opened 2017-09-18
Use real container name to lookup container ip in k8s

pearj:
Related to #211


pearj:
👍

pearj:
lol, it was worth a try, approval validation in progress...

diemol:
Thanks! I'll merge this to the branch I am working now, I plan to implement the integration tests with Minikube next week.

pearj:
Thanks! I attempted to write a test on the bus on the way to work to make codecov happy. But it was looking like it would be fiddly to write a useful test. It does seem like an integration test would be the most useful option.

PR #248, opened 2017-09-19
Zalenium on GKE

laszlocph:
This PR documents how to deploy Zalenium on GKE.


diemol:
Thanks for contributing! We really need help with the k8s docs. I'll merge into a branch I am working on because I want to link it to the main docs.

diemol:
Hi @laszlocph,

I saw that the email you used for your commit is not associated with your account, so you won't show up as a contributor. If you want to, you can just add it to your account :)

image

laszlocph:
Thanks for that @diemol I associated my address :)

If you have specific areas to document in mind, just find me on gitter. It would be very easy to translate the GKE docs to Azure for example, just changing the gcloud commands plus the routing at the end.

diemol:
Cool @laszlocph, I don't have anything specific for now. But that's only because I've written most of it and everything is clear, which might not be the case for a 1st time reader... So if you want to, you can also check what we have and improve it.

Thanks for offering your help!

PR #251, opened 2017-09-21
Release 3.5.3b

### LINKED ISSUE #125: Feature: Allow user to specify a subdirectory for videos, in desiredCapabilities
I was hoping to use this on a Jenkins build agent, where multiple jobs might run various suites. Some of those jobs might run in parallel. I was looking to automate archiving the videos, which are currently successfuly copied by zalenium to /tmp/videos of the host running docker.

But they all get stored in that one directory. At the end of the suite, if I attempt to automate copying and clearing the videos relevant to that suite, I might copy (and clean) videos from a different suite that was running in parallel.

If I were able to set a videoSubdirectory name in the desiredCapabilities, then each suite could have it's videos "organized", if Zalenium were to copy videos to /tmp/videos/$videoSubdirectory/<video file>

Then we can reliably manipulate that specific subdirectory at the end of our suite, without worrying about stepping on any other suite's activity.

I did see that we could set the test name, which gets incorporated into the filename of the video file, but that's not sufficient if I have more than one instance of the suite (with the same tests) running at the same time.

### LINKED ISSUE #151: Filtering videos by test failure status
I was thinking about an enhancement that would be useful to us, and that is being able to filter the list of videos showing only videos that show test failures.

I realise that Zalenium is one step removed from the test framework, however, I was thinking that it should be possible to send a message to Zalenium via the Selenium API that could tell Zalenium the test status.

At first, I thought about creating a completely new selenium command, but then I realised that although Zalenium is a proxy, it can't prevent the calls from being proxied; it can only run before or after the command, it can't decide to respond itself without actually proxying.  Meaning that new commands would make it all the way to the node and they might not like getting commands they don't understand.

So, I started hunting through the selenium API for a command that I could abuse without any side effects, and I think that `org.openqa.selenium.remote.RemoteWebDriver.RemoteWebDriverOptions.addCookie(Cookie)` is the winner.  Because you could basically add an expired cookie from a [TestNG Listener](http://testng.org/doc/documentation-main.html#logging-listeners) inside the onFailure callback with a name of say, `zaleniumTestFailed`, with a value of `true` or something like that.

When we run Selenium tests, we store the WebDriver in a ThreadLocal so that the TestNG Listener can access the WebDriver when tests fail to grab a screenshot (not exactly necessary with video recording anymore).

This could probably further be extended to have a mode where Zalenium can be configured to auto-delete videos that haven't failed or something like that.

I also remember reading in #146 that @elgalu said that there was the possibility of recording videos without compression.  If only failed videos are kept, then maybe uncompressed is a bit more workable, especially if zelenium compresses the videos in the background.

So, would you guys accept a PR that abuses `addCookie` and adds a show only failed videos checkbox to the videos dashboard?

diemol:
Closes #151 and closes #125

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -6,7 +6,7 @@
 
     <groupId>de.zalando.ep</groupId>
     <artifactId>zalenium</artifactId>
-    <version>3.5.3b-SNAPSHOT</version>
+    <version>3.5.3b</version>
     <name>Zalenium</name>
     <description>An on-demand, dynamic local Selenium Grid with an extension to a cloud testing platform.</description>
 
--- a/pom.xml
+++ b/pom.xml
@@ -6,7 +6,7 @@
 
     <groupId>de.zalando.ep</groupId>
     <artifactId>zalenium</artifactId>
-    <version>3.5.3b</version>
+    <version>3.5.3c-SNAPSHOT</version>
     <name>Zalenium</name>
     <description>An on-demand, dynamic local Selenium Grid with an extension to a cloud testing platform.</description>
 


diemol:
👍

elgalu:
:+1:

PR #292, opened 2017-10-29
Cleaning up before running BS tests and upgrading ubuntu base image.

diemol:
Cleaning up before running BS tests and upgrading ubuntu base image.


diemol:
👍

elgalu:
👍

PR #316, opened 2017-11-12
Add content from https://github.com/gswallow/helm-zalenium

gswallow:
This comes out of the discussion in https://github.com/gswallow/helm-zalenium/issues/1.


diemol:
Hi @gswallow,

Thanks for the PR!
I was thinking if it is maybe better to put the files under docs/k8s/helm, because the other folder is used for testing in Travis. Could you please move them there?

Also, if you want to be shown in the contributors, I think you need to add the email used for the commits to your Github account...
image

diemol:
:+1:

PR #358, opened 2017-12-09
Adding initial version of Registry (it is just a copy for now).

diemol:
Adding initial version of Registry (it is just a copy for now).


diemol:
👍

PR #359, opened 2017-12-10
Retrying registration + removing overhead in proxy removal

diemol:
Retrying registration + removing overhead in proxy removal


diemol:
👍

PR #386, opened 2018-01-01
Adding how it works and contributing section for new docs

diemol:
Adding how it works and contributing section for new docs


diemol:
👍

Issue #382, opened 2017-12-28
live preview doesn't work with --net=host


boakley:
When I use `--net host`, the previewer (http://localhost:4444/grid/admin/live) doesn't work. It shows "Bad Gateway". When I change the command line to use `--net bridge`, the preview shows.

Even though the preview doesn't work, the test passes and a video is properly created and can be viewed from the dashboard once the test finishes. 

Zalenium Image Version(s):  3.8.,1

Docker Version:  17.04.0-ce, build 4845c56

If using docker-compose, version:  

OS: LinuxMint 18

Docker Command to start Zalenium:  

<pre>docker run --net=host --rm -ti --name zalenium -p 4444:4444  -v /var/run/docker.sock:/var/run/docker.sock  -v /tmp/videos:/home/seluser/videos --privileged dosel/zalenium start --timeZone "America/Chicago" 
</pre>

When I use the above, the previewer shows "bad gateway". If I change `--net=host` to `--net=bridge`, the previewer works (but my tests don't, since they are referencing localhost)

If using Kubernetes, specify your environment, and if relevant your manifests:  N/A

## Expected Behavior -

The live preview should work

## Actual Behavior -

The live preview does not work. Instead, I see this image:

![zalenium_bad_gateway](https://user-images.githubusercontent.com/1123938/34422447-3b84a936-ebdb-11e7-93cd-761daa7937a8.png)

diemol:
Hi @boakley,

Sorry for the late reply, we were on holidays.

I think it is related to the latest change, I'll double check tomorrow.

diemol:
Hi, I just merged #392, which fixes it, I'll fix another issue between today and Friday and then make a release. Thanks for reporting this.

diemol:
Hi @boakley,

This was just released, please try it with the latest tag https://github.com/zalando/zalenium/releases/tag/3.8.1f

Feel free to reopen in case something related is not working.

boakley:
It seems to be working much better. I've made a couple dozen test runs, and
only during one of those did I get the "502 Bad Gateway" display.
Re-running the test and refreshing the page cleared it up. I haven't been
able to reproduce it a second time. If I'm able to reliably reproduce it,
I'll let you know.

diemol:
Normally the "502 Bad Gateway" can happen when the noVNC does not start properly in the container, and this could be every now and then depending on the load. 
That is different from what you reported, because it was just an issue how the proxying was handled. 

Thanks for the quick feedback!

[...]

PR #392, opened 2018-01-03
Simple change to make live preview work with --net=host #382

diemol:
Simple change to make live preview work with --net=host #382


diemol:
👍

Issue #332, opened 2017-11-20
Passing additional options to sauce connect


ryudice:
Is there a way to pass additional command line arguments to Sauce Connect? I'm having issues with resolving internal DNS domains.

diemol:
Hi @ryudice,

Not really, but in theory any needed parameter could be passed to the container via an env var and then mapped to the binary when starting it. This [file](https://github.com/zalando/zalenium/blob/master/scripts/start-saucelabs.sh) is the one starting it.

What parameter do you have in mind?

ryudice:
Hi @diemol, the DNS parameter:
![image](https://user-images.githubusercontent.com/125211/33030490-09187bde-cde1-11e7-9934-de9d207a83b0.png)

diemol:
OK, this should not be that complicated to add.
Do you feel like adding it via a PR? I could tell you where to get started and how to test it.

Otherwise I can take a look at it in 4-6 weeks since we are working in different stuff now.

diemol:
Hi @ryudice,

This was just implemented in #393, we'll release it tomorrow so you can use it.

You'd use it like this (note the **SAUCE_TUNNEL_DNS** var):

```
docker run --rm -ti --name zalenium -p 4444:4444 \
    -v /var/run/docker.sock:/var/run/docker.sock \
    -v /tmp/videos:/home/seluser/videos -e SAUCE_USERNAME -e SAUCE_ACCESS_KEY \
    -e SAUCE_TUNNEL_DNS="8.8.8.8,8.8.4.4:53" \
    --privileged dosel/zalenium start --sauceLabsEnabled true --startTunnel true
```

diemol:
Hi @ryudice,

This was just released, please try it with the latest tag https://github.com/zalando/zalenium/releases/tag/3.8.1f

Feel free to reopen in case something related is not working. 

PS: Don't forget to star the repo to support us!

PR #393, opened 2018-01-04
Adding option to pass a dns value to Sauce Connect #332

diemol:
Adding option to pass a dns value to Sauce Connect #332


diemol:
👍

Issue #396, opened 2018-01-08
Rancher compatibility


monkpit:
## Expected Behavior -
I should be able to run zalenium via a docker-compose file without manually doing "docker pull eglalu/selenium" beforehand and having the zalenium able to "see" the image existing on the host.

## Actual Behavior -
Error messages during `start`:

```
Ensuring docker-selenium is available...
1/8/2018 3:18:35 PMSeems that docker-selenium's image has not been pulled yet
1/8/2018 3:18:35 PMPlease run 'docker pull elgalu/selenium', or use your own compatible image via --seleniumImageName
```

If `eglau/selenium` is required for the docker-compose file to work, then it should be a part of the docker-compose file itself and managed by docker that way. What is the workaround? Can this be done via docker networking?

My use case is: trying to run zalenium via [Rancher](https://rancher.com/).

diemol:
Hi @monkpit,

This was just implemented, and it will be released between today and tomorrow. You'll be able to use it like this:

```
docker run --rm -ti --name zalenium -p 4444:4444 \
    -v /var/run/docker.sock:/var/run/docker.sock \
    -v /tmp/videos:/home/seluser/videos -e PULL_SELENIUM_IMAGE="true" \
    --privileged zalenium:3.8.1g-SNAPSHOT start
```

I'll comment here when the release is done.

monkpit:
@diemol Awesome! Thanks for the reply!

diemol:
Hi again,

This was just released, please check https://github.com/zalando/zalenium/releases/tag/3.8.1g

Let us know if things work as expected. I'll close this issue, but feel free to reopen it in case something didn't work.

monkpit:
Hmm - it works and the selenium containers get created, but the zalenium containers shows timeout errors because it was unable to connect to them.

Zalenium shows this:

```

1/9/2018 3:08:25 PM2018-01-09 22:08:25 INFO  session:149 - Scavenging every 600000ms
1/9/2018 3:08:25 PM2018-01-09 22:08:25 INFO  ContextHandler:788 - Started o.s.j.s.ServletContextHandler@475e586c{/,null,AVAILABLE}
1/9/2018 3:08:25 PM2018-01-09 22:08:25 INFO  AbstractConnector:286 - Started ServerConnector@1460a8c0{HTTP/1.1,[http/1.1]}{0.0.0.0:4445}
[...]

PR #399, opened 2018-01-09
Adding flag to pull docker-selenium image in case it is not present.

diemol:
Adding flag to pull docker-selenium image in case it is not present. #396


diemol:
👍

PR #406, opened 2018-01-18
Upgrading docker versions and docker-client dependency

diemol:
Upgrading docker versions and docker-client dependency


diemol:
👍

Issue #414, opened 2018-01-23
Cannot retrieve videos from BrowserStack


jbengit:
Zalenium Image Version(s):  
3.8.1h
Docker Version:  
17.09.1-ce
OS: 
Ubuntu 14
Docker Command to start Zalenium:  
 curl -N -sSL https://raw.githubusercontent.com/dosel/t/i/p  | bash  -s start     --sendAnonymousUsageInfo false --videoRecordingEnabled true 

## Expected Behavior -
I am using BrowserStack through Zalenium, with real device testing, and expect to see the test video in the Zalenium dashboard.

## Actual Behavior -
At the end of the test, Zalenium does not manage to retrieve the video from BrowserStack, here is the exception that I found in the logs:
`2018-01-22 17:09:32 SEVERE java.lang.UnsupportedOperationException: JsonNull
java.lang.UnsupportedOperationException: JsonNull
        at com.google.gson.JsonElement.getAsString(JsonElement.java:192)
        at de.zalando.ep.zalenium.proxy.BrowserStackRemoteProxy.getTestInformation(BrowserStackRemoteProxy.java:90)
        at de.zalando.ep.zalenium.proxy.CloudTestingRemoteProxy.lambda$addTestToDashboard$2(CloudTestingRemoteProxy.java:236)
        at java.lang.Thread.run(Thread.java:748)`

It occurs because the BrowserStack API is returning null values for some specific fields (**'browser' & 'browser_version'**):
`{"automation_session":{..."os":"ios","os_version":"11.0","browser_version":null,"browser":null,"device":"iPhone 8","status":"done"...}}`

Thank you

diemol:
Hi @jbengit,

Thanks for the clear error message, I'll add some NPE validations and release again.

jbengit:
Thank you for your awesome work!

jbengit:
It's working now thank you!

diemol:
Great!

PR #419, opened 2018-01-27
Adding JSONNull validations and updating Spotify Client

diemol:
Adding JSONNull validations and updating Spotify Client #414


diemol:
👍

PR #421, opened 2018-01-28
Extending the PULL_SELENIUM_IMAGE to actually pull the image

diemol:
in case it is not present before creating a container #396


diemol:
👍

PR #464, opened 2018-03-05
Setting amount of executed tests to maximum allowed when the node is …

### LINKED ISSUE #426: parallel tests: WebDriverException Connection refused and Error forwarding the request happening with greater frequency
Zalenium Image Version(s):  latest -- pull is automated before every suite launch.
Docker Version:  17.04.0-ce and 17.12.0-ce
OS: SLES 12 and Ubuntu 16.04

I run tests in parallel -- there is 1 master zalenium container that spins up 6 grid containers, with th chromeContainers 6 option.

This works fine in general, and has been very reliable for the last month or two.

Over the past week, I'll see 2-3 tests fail (across ~140 total in the suite), with various webDriverException errors regarding connection refused, error forwarding the request, and something like an "INVALID_SESSION" error, without changing things here.

I don't really know where to begin.

Let me ask this -- is there some kind of delay needed before starting a test, when the previous test has just finished?

This seems to happen mid-suite when things are busy. Let's say 6 chromeContainers are each running a test, recording video, etc. Let's say container number 3 is the first to finish its test. The last step of that test is to close the browser. Then, the next test starts, and very quickly attempts to instantiate a new remoteWebDriver, and that request gets sent to the 'master' Zalenium container, acting as the selenium hub. What happens if the child grid container is still wrapping up re: copying the video file, or any other kind of cleanup? Or is there any other kind of race condition that can lead to this occurring? I'm often seeing 'connection refused' in the exception.

On a related note, if there are 6 chromeContainers, all running a test.... what happens if a 7th request comes in? Does it queue? Not sure what the expected behavior is.

This is becoming painful enough that we may need to stop using -- we have people that need to triage each of these failures to ensure it isn't a real regression.

Sometimes we don't see any problem, and when we do see a problem, it's always a different test. It looks to be some kind of a race condition, and it seems to have got worse over the last week or so, where I hadn't seen this occur other than once or twice over the past month or two.

diemol:
…shutting down, fixes #426

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,9 +1,9 @@
 #== Ubuntu xenial is 16.04, i.e. FROM ubuntu:16.04
 # Find latest images at https://hub.docker.com/r/library/ubuntu/
 # Layer size: big: ~130 MB
-FROM ubuntu:xenial-20180123
+FROM ubuntu:xenial-20180228
 ENV UBUNTU_FLAVOR="xenial" \
-    UBUNTU_DATE="20180123"
+    UBUNTU_DATE="20180228"
 
 #== Ubuntu flavors - common
 RUN  echo "deb http://archive.ubuntu.com/ubuntu ${UBUNTU_FLAVOR} main universe\n" > /etc/apt/sources.list \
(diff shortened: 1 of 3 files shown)


diemol:
👍

diemol:
👍

thilp:
:+1:

PR #467, opened 2018-03-08
Displaying only one message instead of cascading them, fixes #450

### LINKED ISSUE #450: Cascading zalenium messages
I am using zalenium messages by adding a cookie and it works beautifully.

Is it possible to remove the first zalenium message before displaying the 2nd zalenium message?

image

In the above screenshot I would like to remove: 

"Starting test: SC Munki is loading properly"

before displaying

"Starting test: Checking title"

diemol:
Displaying only one message instead of cascading them, fixes #450


diemol:
👍

PR #478, opened 2018-03-12
Tuning capability matchers and validations, fixes #457

### LINKED ISSUE #457: Browserstack: Problems related to wrong capabilities
Zalenium Image Version(s):  
3.9.1a 

Docker Version:  
Running in openshift origin 3.7

## Expected Behavior -
When browserstack is enabled in Zalenium and your test requests capabilities that are missing some mandatory capability, this should be handled or at least responded to the client somehow

## Actual Behavior -
This is the message shown in the output of zalenium if your test don't have the `browserName` capability for instance:
```
09:25:42.331 WARN - Capability {browserstack.local=true, browserstack.localIdentifier=zalenium, acceptSslCerts=true, browserstack.video=true, se:CONFIG_UUID=9bd3cbc7-5f1b-4f94-80e3-a6ab88b10b9b, os_version=7.0, javascriptEnabled=true, real_mobile=true, zal:name=pytest android, device=Samsung Galaxy S8} does no contain browserName key.
```
This message is repeated (because I guess Zalenium just keeps trying) until some sort of timeout occur against browserstack (not sure about this though).

diemol:
Allows to forward tests in mobile devices to the cloud providers.

### DIFF
--- a/src/main/java/de/zalando/ep/zalenium/matcher/DockerSeleniumCapabilityMatcher.java
+++ b/src/main/java/de/zalando/ep/zalenium/matcher/DockerSeleniumCapabilityMatcher.java
@@ -42,6 +42,12 @@ public class DockerSeleniumCapabilityMatcher extends DefaultCapabilityMatcher {
         logger.log(Level.FINE, ()-> String.format("Validating %s in node with capabilities %s", requestedCapability,
                 nodeCapability));
 
+        if (!requestedCapability.containsKey(CapabilityType.BROWSER_NAME)) {
+            logger.log(Level.FINE, () -> String.format("%s Capability %s does no contain %s key, a docker-selenium " +
+                    "node cannot be started without it", proxy.getId(), requestedCapability, CapabilityType.BROWSER_NAME));
+            return false;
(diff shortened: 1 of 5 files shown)


src/main/java/de/zalando/ep/zalenium/matcher/DockerSeleniumCapabilityMatcher.java
```diff
@@ -42,6 +42,12 @@ public boolean matches(Map<String, Object> nodeCapability, Map<String, Object> r
         logger.log(Level.FINE, ()-> String.format("Validating %s in node with capabilities %s", requestedCapability,
                 nodeCapability));
 
+        if (!requestedCapability.containsKey(CapabilityType.BROWSER_NAME)) {
+            logger.log(Level.FINE, () -> String.format("%s Capability %s does no contain %s key, a docker-selenium " +
```
peerster: Small typo.. I would replace `does no` to `doesn't`

diemol:
👍

peerster:
:+1:

Issue #441, opened 2018-02-15
Pod creation flood when running on k8s


Vlaaaaaaad:
Issue created after discussion on [Slack](https://seleniumhq.slack.com/archives/C99KY79EK/p1518689803000467) and Gitter.

### Expected Behavior

When running a Selenium test on Kubernetes that requests 300 instances of Firefox( several scenarios running in parallel), I expected 300 pods to be created to serve that.

### Actual Behavior

Significantly more pods are are created to serve the request. The `maxDockerSeleniumContainers` limit was set to 600 and it was hit( the limit seemed to work and Zalenium reported stopping the requests due to the limit).

The following message can be seen in the logs for Zalenium hub:
```
INFO - [DS] Request has waited 30 attempts for a node, something went wrong with the previous attempts, creating a new node for { ... }
```

### Investigation and write-up

Zalenium is very... eager to schedule pods. When running a test that requests 10 pods, k8s would schedule 4, have no more space, cluster-autoscaler would create a new node that would join the cluster in a couple minutes and then the 6 pending pods would be scheduled. Meanwhile Zalenium would keep requesting new 6 pods over and over again. This leads to 20-something pods for a test that requested 10.

To confirm this an over-powered cluster was created where there were no issues with the 10-pod scheduling and subsequent cleanup. There were 10 pods requested, 10 were created, tests ran, pods were deleted. Perfect.

The same problem comes back when requesting more pods. At 300 pods, even on a larger cluster, scheduling and launching all the pods takes a while and Zalenium keeps requesting pods. For a request of 300 pods, Zalenium tries to create more than 600 pods. 

Looking a bit into the code points to this being hardcoded at
https://github.com/zalando/zalenium/blob/09f22a4375fd3f567303ec2017be7f31ba58bede/src/main/java/de/zalando/ep/zalenium/proxy/DockerSeleniumStarterRemoteProxy.java#L374 

I see no way to increase a timeout or the 30 attempts value. Something like "_OK Zalenium, you requested pods, just wait like 2 minutes for them to be ready_" would be a great addition.

### Env
**Zalenium Image Version**: 3

**Kubernetes Version**: 1.8.6, cluster created by kops, running on AWS

**Command to start Zalenium**:  installed on the cluster by deploying the helm chart with the following `values.yaml`:
```
hub:
  ## The repository and image
  ## ref: https://hub.docker.com/r/selenium/hub/
  image: "dosel/zalenium"

  ## The tag for the image
[...]

PR #487, opened 2018-03-19
WIP for #441

diemol:
WIP for #441 
Versions upgrade
Added extra logging
Removing old session requests


src/main/java/de/zalando/ep/zalenium/registry/ZaleniumRegistry.java
```diff
@@ -55,6 +58,8 @@ public ZaleniumRegistry(Hub hub) {
         this.newSessionQueue = new NewSessionRequestQueue();
         proxies = new ProxySet((hub != null) ? hub.getConfiguration().throwOnCapabilityNotPresent : true);
         this.matcherThread.setUncaughtExceptionHandler(new UncaughtExceptionHandler());
+        Environment environment = new Environment();
+        maxRequestAge = environment.getIntEnvVariable("ZALENIUM_MAX_REQUEST_AGE", 5);
```
pearj: I think seconds will be better here than minutes, because otherwise this is saying the shortest timeout you can have is 1 minute, but sometimes you might want to fail things earlier.

diemol: Sounds good, just changed it

diemol:
👍

elgalu:
👍

PR #507, opened 2018-04-02
Allow parameters to be set via environment variables.

pearj:
This is especially useful for OpenShift where the UI easily allows you
to edit environment variables, but command arguments are more fiddly to
deal with.


diemol:
Thanks @pearj!
This makes sense, we only have the command line parameters because the first version was just an executable. Hopefully in the future we can only have env vars.

elgalu:
:+1:

diemol:
👍

PR #509, opened 2018-04-02
Adding support for storage opt parameter in docker.

### LINKED ISSUE #505: container /dev/dm-*  elgalu/selenium base size 25G ?
Hi.,

After we start zalenium docker which in turns boot up  containers using image name -  elgalu/selenium. Question - how can we increase the default storage size of container - 25G to 50G ?  I tried adding  --storage-opt size=50G , it only increased storage size of zalenium container , however container of elgalu/selenium did not adjust to 50G, still 25G.

Please advice how do we increase ?

diemol:
Fixes #505


elgalu:
:+1:

diemol:
👍

PR #527, opened 2018-04-16
Request for Comment - Move starter proxy responsibilities into proxy set.

dspasojevic:
### Description
Removes the remote starter proxy and moves its responsibilities into a custom ProxySet.

This PR is for comment only - I don't expect it to be merged in current incarnation.

### Motivation and Context
Permits closer tracking of starting, started and closing containers / proxies. Prevents creation of duplicates and orphan containers.

### How Has This Been Tested?
Tested in an OpenShift environment. Requires *more* testing.

### Types of changes
- [X] Breaking change (fix or feature that would cause existing functionality to change)

### Checklist:
- [X] My change requires a change to the documentation.


dspasojevic:
@pearj do you mind taking a quick look at this change?

scripts/zalenium.sh
```diff
@@ -21,6 +21,7 @@ START_TUNNEL=${START_TUNNEL:-false}
 DEBUG_ENABLED=${DEBUG_ENABLED:-false}
 KEEP_ONLY_FAILED_TESTS=${KEEP_ONLY_FAILED_TESTS:-false}
 LOG_JSON=${LOG_JSON:-false}
+NEW_SESSION_WAIT_TIMEOUT=${NEW_SESSION_WAIT_TIMEOUT:-300000}
```
peerster: Duplicate line of row 26

src/main/java/de/zalando/ep/zalenium/proxy/DockerSeleniumRemoteProxy.java
```diff
@@ -41,8 +41,12 @@
...
 import java.util.Map;
 import java.util.Optional;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.concurrent.atomic.AtomicLong;
+import java.util.logging.Level;
+import java.util.logging.Logger;
```
peerster: It think we want to use slf4j. @dennisgranath recently made changes to start using it instead of java.util.

pearj: Yeah that's true

scripts/zalenium.sh
dspasojevic: Fixed.

src/main/java/de/zalando/ep/zalenium/proxy/DockerSeleniumRemoteProxy.java
dspasojevic: Will change.

src/test/java/de/zalando/ep/zalenium/proxy/SauceLabsRemoteProxyTest.java
```diff
@@ -44,6 +44,7 @@
 
 public class SauceLabsRemoteProxyTest {
 
+	private DockerSeleniumRemoteProxy proxy;
```
peerster: Wrong indentation it seems :)

dennisgranath:
When starting zalenium on docker using 
`docker run --rm -ti --name zalenium -p 4444:4444         -v /var/run/docker.sock:/var/run/docker.sock         -v /tmp/videos:/home/seluser/videos --privileged zalenium-pr start`

I get following failure when starting up the second container: 

`10:08:48.421 [AutoStartProxyPoolPoller] INFO  d.z.e.z.proxy.AutoStartProxySet - Failed to start container.
10:08:49.502 [AutoStartProxyPoolPoller] INFO  d.z.e.z.proxy.AutoStartProxySet - Autostarting container, because 1 is less than min 2
10:08:49.528 [AutoStartProxyPoolPoller] WARN  d.z.e.z.c.DockerContainerClient - null Error while starting a new container
com.spotify.docker.client.exceptions.DockerRequestException: Request error: POST unix://localhost:80/containers/create?name=zalenium_40000: 409, body: {"message":"Conflict. The container name \"/zalenium_40000\" is already in use by container \"7622621076a2da00a42ca7698ecb5f043ca5a73ae596cfc7df05c11b53f20c8b\". You have to remove (or rename) that container to be able to reuse that name."}`

docker/Dockerfile
```diff
@@ -318,11 +318,7 @@ RUN cd /tmp \
 # TestingBot Tunneling #
 # -----------------------#
 # https://testingbot.com/support/other/tunnel
-ENV TB_TUNNEL_URL="https://testingbot.com/tunnel/testingbot-tunnel.jar"
```
pearj: Probably should leave this in there. You just removed it because we doesn't work at the client site right?

dspasojevic:
@dennisgranath I've updated the `DockerContainerClient` to randomise six chars in the same way that kubernetes does and implemented the `isTerminated` check. It spins containers up for me locally. I don't have selenium tests that I can run where I am now. It would be great if you could check whether it works for you.

docker/Dockerfile
dspasojevic: Yep. I'll revert this change.

dennisgranath:
Thanks @dspasojevic I will run them a soon as I can. Might be able to do it later today.

docker/Dockerfile
pearj: I think you could probably add: `ARG testingBotEnabled=true` on the line beforehand. Then you could add `[ "${testingBotEnabled}" == "true" ] && \` before the wget line

src/main/java/de/zalando/ep/zalenium/matcher/DockerSeleniumCapabilityMatcher.java
```diff
@@ -48,22 +51,6 @@ public boolean matches(Map<String, Object> nodeCapability, Map<String, Object> r
             return false;
         }
 
-        // We do this because the starter node does not have the browser versions when Zalenium starts
```
pearj: @diemol Do you know if this code is necessary to not?

src/main/java/de/zalando/ep/zalenium/matcher/ZaleniumCapabilityMatcher.java
```diff
@@ -31,7 +32,7 @@ public boolean matches(Map<String, Object> nodeCapability, Map<String, Object> r
                 nodeCapability));
 
         for (RemoteProxy remoteProxy : proxy.getRegistry().getAllProxies()) {
-            if ((remoteProxy instanceof DockerSeleniumStarterRemoteProxy) &&
+            if ((remoteProxy instanceof DockerSeleniumRemoteProxy) &&
```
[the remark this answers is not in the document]
dspasojevic: I *think* this is what must be duplicated here: https://github.com/zalando/zalenium/blob/master/src/main/java/de/zalando/ep/zalenium/proxy/DockerSeleniumStarterRemoteProxy.java#L189

dennisgranath:
Ok now it seems to work when using plain docker however I get following exception when running my selenium tests (the selenium tests are passing though): 

`ded396482fb124eef3a0bd3c5f661051fa8f0c7]] INFO  d.z.e.z.c.DockerContainerClient - http://172.17.0.3:40000 [bash, -c, cleanup-container.sh]
17:03:27.519 [AutoStartProxyPoolPoller] INFO  d.z.e.z.p.DockerSeleniumRemoteProxy - [c0f7d009e72e9a8c9a2a46a27ded396482fb124eef3a0bd3c5f661051fa8f0c7] is idle.
17:03:27.519 [AutoStartProxyPoolPoller] INFO  d.z.e.z.p.DockerSeleniumRemoteProxy - c0f7d009e72e9a8c9a2a46a27ded396482fb124eef3a0bd3c5f661051fa8f0c7 Shutting down node due to proxy being idle after test.
17:03:31.262 [pool-3-thread-2] INFO  o.o.g.s.proxy.DefaultRemoteProxy - c0f7d009e72e9a8c9a2a46a27ded396482fb124eef3a0bd3c5f661051fa8f0c7 Marking the node as down because it was stopped after 1 tests.
17:03:31.262 [pool-3-thread-2] WARN  d.z.e.z.registry.ZaleniumRegistry - Cleaning up stale test sessions on the unregistered node http://172.17.0.3:40000
17:03:31.291 [pool-3-thread-2] WARN  d.z.e.z.c.DockerContainerClient - null Error while stopping the container
com.spotify.docker.client.exceptions.ContainerNotFoundException: Container not found: c0f7d009e72e9a8c9a2a46a27ded396482fb124eef3a0bd3c5f661051fa8f0c7
	at com.spotify.docker.client.DefaultDockerClient.stopContainer(DefaultDockerClient.java:772)
	at de.zalando.ep.zalenium.container.DockerContainerClient.stopContainer(DockerContainerClient.java:162)
	at de.zalando.ep.zalenium.proxy.DockeredSeleniumStarter.stopContainer(DockeredSeleniumStarter.java:303)
	at de.zalando.ep.zalenium.proxy.AutoStartProxySet.remove(AutoStartProxySet.java:199)
	at de.zalando.ep.zalenium.registry.ZaleniumRegistry.removeIfPresent(ZaleniumRegistry.java:161)
	at org.openqa.grid.selenium.proxy.DefaultRemoteProxy.onEvent(DefaultRemoteProxy.java:193)
	at org.openqa.grid.selenium.proxy.DefaultRemoteProxy.addNewEvent(DefaultRemoteProxy.java:179)
	at de.zalando.ep.zalenium.proxy.DockerSeleniumRemoteProxy.shutdownNode(DockerSeleniumRemoteProxy.java:702)
	at de.zalando.ep.zalenium.proxy.DockerSeleniumRemoteProxy$1.run(DockerSeleniumRemoteProxy.java:468)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
	at java.lang.Thread.run(Thread.java:748)
Caused by: com.spotify.docker.client.exceptions.DockerRequestException: Request error: POST unix://localhost:80/containers/c0f7d009e72e9a8c9a2a46a27ded396482fb124eef3a0bd3c5f661051fa8f0c7/stop?t=5: 404, body: {"message":"No such container: c0f7d009e72e9a8c9a2a46a27ded396482fb124eef3a0bd3c5f661051fa8f0c7"}

	at com.spotify.docker.client.DefaultDockerClient.propagate(DefaultDockerClient.java:2702)
	at com.spotify.docker.client.DefaultDockerClient.request(DefaultDockerClient.java:2663)
	at com.spotify.docker.client.DefaultDockerClient.stopContainer(DefaultDockerClient.java:766)
	... 11 common frames omitted
Caused by: javax.ws.rs.NotFoundException: HTTP 404 Not Found
	at org.glassfish.jersey.client.JerseyInvocation.convertToException(JerseyInvocation.java:1008)
	at org.glassfish.jersey.client.JerseyInvocation.translate(JerseyInvocation.java:816)
	at org.glassfish.jersey.client.JerseyInvocation.access$700(JerseyInvocation.java:92)
	at org.glassfish.jersey.client.JerseyInvocation$5.completed(JerseyInvocation.java:773)
	at org.glassfish.jersey.client.ClientRuntime.processResponse(ClientRuntime.java:198)
	at org.glassfish.jersey.client.ClientRuntime.access$300(ClientRuntime.java:79)
	at org.glassfish.jersey.client.ClientRuntime$2.run(ClientRuntime.java:180)
	at org.glassfish.jersey.internal.Errors$1.call(Errors.java:271)
	at org.glassfish.jersey.internal.Errors$1.call(Errors.java:267)
	at org.glassfish.jersey.internal.Errors.process(Errors.java:315)
	at org.glassfish.jersey.internal.Errors.process(Errors.java:297)
	at org.glassfish.jersey.internal.Errors.process(Errors.java:267)
	at org.glassfish.jersey.process.internal.RequestScope.runInScope(RequestScope.java:340)
	at org.glassfish.jersey.client.ClientRuntime$3.run(ClientRuntime.java:210)
	at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511)
	at java.util.concurrent.FutureTask.run(FutureTask.java:266)
	... 3 common frames omitted
17:03:31.542 [AutoStartProxyPoolPoller] INFO  d.z.e.z.c.DockerContainerClient - Container ContainerCreationStatus [isCreated=true, containerName=zalenium_40000_oDTbUA, containerId=c0f7d009e72e9a8c9a2a46a27ded396482fb124eef3a0bd3c5f661051fa8f0c7, nodePort=40000] not found - it is terminated.
17:03:31.542 [AutoStartProxyPoolPoller] INFO  d.z.e.z.proxy.AutoStartProxySet - Container ContainerCreationStatus [isCreated=true, containerName=zalenium_40000_oDTbUA, containerId=c0f7d009e72e9a8c9a2a46a27ded396482fb124eef3a0bd3c5f661051fa8f0c7, nodePort=40000] is terminated. Removing from tracked set.
17:03:51.587 [AutoStartProxyPoolPoller] INFO  d.z.e.z.proxy.AutoStartProxySet - Status:
┌────┬───────┬─────────┬─────────┬───────────┬──────────────┬──────┬───────────┬─────────────┬───────────┐
│ Id │ Proxy │ Created │ Started │ Last Used │ Last Session │ Busy │ Timed Out │ Terminating │ Tests Run │
├┬┬┬┬┼┬┬┬┬┬┬┬┼┬┬┬┬┬┬┬┬┬┼┬┬┬┬┬┬┬┬┬┼┬┬┬┬┬┬┬┬┬┬┬┼┬┬┬┬┬┬┬┬┬┬┬┬┬┬┼┬┬┬┬┬┬┼┬┬┬┬┬┬┬┬┬┬┬┼┬┬┬┬┬┬┬┬┬┬┬┬┬┼┬┬┬┬┬┬┬┬┬┬┬┤
└┴┴┴┴┴┴┴┴┴┴┴┴┴┴┴┴┴┴┴┴┴┴┴┴┴┴┴┴┴┴┴┴┴┴┴┴┴┴┴┴┴┴┴┴┴┴┴┴┴┴┴┴┴┴┴┴┴┴┴┴┴┴┴┴┴┴┴┴┴┴┴┴┴┴┴┴┴┴┴┴┴┴┴┴┴┴┴┴┴┴┴┴┴┴┴┴┴┴┴┴┴┴┴┴┘`

dspasojevic:
@dennisgranath thanks for testing again. It looks like I just need to handle that exception while shutting down the container. Part of the change is it to always attempt to shutdown the container when the proxy is removed to ensure that orphan, untracked containers aren't left around. I think that zalenium itself would have been fine, and would have spun up another container if necessary.

dspasojevic:
@dennisgranath I've pushed up a change to specifically handle the `ContainerNotFoundException`. It would be great if you could try again.

dennisgranath:
Ok at first we have 2 containers ready to serve requests:
```
┌───────────────────────┬─────────────────────────┬─────────────────────────┬─────────────────────────┬─────────────────────┬─────────────────────────┬───────┬───────────┬─────────────┬───────────┐
│ Id                    │ Proxy                   │ Created                 │ Started                 │ Last Used           │ Last Session            │  Busy │ Timed Out │ Terminating │ Tests Run │
├───────────────────────┼─────────────────────────┼─────────────────────────┼─────────────────────────┼─────────────────────┼─────────────────────────┼───────┼───────────┼─────────────┼───────────┤
│ zalenium_40000_xsW02C │ http://172.17.0.4:40000 │ 2018-04-18T08:28:13.281 │ 2018-04-18T08:28:19.102 │ 1970-01-01T01:00:00 │ 1970-01-01T00:59:59.999 │ false │     false │       false │         0 │
│ zalenium_40000_ubvN5A │ http://172.17.0.3:40000 │ 2018-04-18T08:28:12.859 │ 2018-04-18T08:28:18.065 │ 1970-01-01T01:00:00 │ 1970-01-01T00:59:59.999 │ false │     false │       false │         0 │
└───────────────────────┴─────────────────────────┴─────────────────────────┴─────────────────────────┴─────────────────────┴─────────────────────────┴───────┴───────────┴─────────────┴───────────┘
```

I run my selenium tests and we have:
```
┌───────────────────────┬─────────────────────────┬─────────────────────────┬─────────────────────────┬─────────────────────────┬─────────────────────────┬───────┬───────────┬─────────────┬───────────┐
│ Id                    │ Proxy                   │ Created                 │ Started                 │ Last Used               │ Last Session            │  Busy │ Timed Out │ Terminating │ Tests Run │
├───────────────────────┼─────────────────────────┼─────────────────────────┼─────────────────────────┼─────────────────────────┼─────────────────────────┼───────┼───────────┼─────────────┼───────────┤
│ zalenium_40000_xsW02C │ http://172.17.0.4:40000 │ 2018-04-18T08:28:13.281 │ 2018-04-18T08:28:19.102 │ 1970-01-01T01:00:00     │ 1970-01-01T00:59:59.999 │ false │     false │       false │         0 │
│ zalenium_40000_ubvN5A │ http://172.17.0.3:40000 │ 2018-04-18T08:28:12.859 │ 2018-04-18T08:28:18.065 │ 2018-04-18T08:29:04.069 │ 2018-04-18T08:28:51.563 │ false │     false │       false │         1 │
└───────────────────────┴─────────────────────────┴─────────────────────────┴─────────────────────────┴─────────────────────────┴─────────────────────────┴───────┴───────────┴─────────────┴───────────┘
```

I run one more time and get:
```
┌───────────────────────┬─────────────────────────┬─────────────────────────┬─────────────────────────┬─────────────────────────┬─────────────────────────┬───────┬───────────┬─────────────┬───────────┐
│ Id                    │ Proxy                   │ Created                 │ Started                 │ Last Used               │ Last Session            │  Busy │ Timed Out │ Terminating │ Tests Run │
├───────────────────────┼─────────────────────────┼─────────────────────────┼─────────────────────────┼─────────────────────────┼─────────────────────────┼───────┼───────────┼─────────────┼───────────┤
│ zalenium_40000_xsW02C │ http://172.17.0.4:40000 │ 2018-04-18T08:28:13.281 │ 2018-04-18T08:28:19.102 │ 2018-04-18T08:34:57.946 │ 2018-04-18T08:34:45.958 │ false │     false │       false │         1 │
│ zalenium_40000_ubvN5A │ http://172.17.0.3:40000 │ 2018-04-18T08:28:12.859 │ 2018-04-18T08:28:18.065 │ 2018-04-18T08:29:04.069 │ 2018-04-18T08:28:51.563 │ false │     false │       false │         1 │
└───────────────────────┴─────────────────────────┴─────────────────────────┴─────────────────────────┴─────────────────────────┴─────────────────────────┴───────┴───────────┴─────────────┴───────────┘
```
And another run:
```
┌───────────────────────┬─────────────────────────┬─────────────────────────┬─────────────────────────┬─────────────────────────┬─────────────────────────┬───────┬───────────┬─────────────┬───────────┐
│ Id                    │ Proxy                   │ Created                 │ Started                 │ Last Used               │ Last Session            │  Busy │ Timed Out │ Terminating │ Tests Run │
├───────────────────────┼─────────────────────────┼─────────────────────────┼─────────────────────────┼─────────────────────────┼─────────────────────────┼───────┼───────────┼─────────────┼───────────┤
│ zalenium_40000_fA6lO7 │ http://172.17.0.5:40000 │ 2018-04-18T08:36:28.763 │ 2018-04-18T08:36:32.922 │ 2018-04-18T08:36:45.489 │ 2018-04-18T08:36:33.301 │  true │     false │       false │         1 │
│ zalenium_40000_ubvN5A │ http://172.17.0.3:40000 │ 2018-04-18T08:28:12.859 │ 2018-04-18T08:28:18.065 │ 2018-04-18T08:29:04.069 │ 2018-04-18T08:28:51.563 │ false │     false │       false │         1 │
└───────────────────────┴─────────────────────────┴─────────────────────────┴─────────────────────────┴─────────────────────────┴─────────────────────────┴───────┴───────────┴─────────────┴───────────┘
```
So in the last run it actually removed the already used container (zalenium_40000_xsW02C) and created a new one (zalenium_40000_fA6lO7). Is this behaviour expected? I guess since zalenium_40000_xsW02C could not be reused it could remove it directly and create a new one to serve more request. Have not tried on Openshift yet but will try to do that during the day.

dspasojevic:
@dennisgranath containers will be started and shutdown depending on the min and max size, whether containers should be reused, the max test count and the idle timeout.

It should always try to keep min (desired) containers ready. 

I would expect that if a container was removed and your min is set to 2, then a new container would be created immediately.

If you attach the full log, I’ll take a look.

dennisgranath:
@peerster Browserstack integration is working fine!

dennisgranath:
@dspasojevic [zalenium.log](https://github.com/zalando/zalenium/files/1923285/zalenium.log)

dspasojevic:
@dennisgranath here are the logs showing it shutting down (logs the container id which is not as friendly as the name):
```
11:25:20.376 [AutoStartProxyPoolPoller] INFO  d.z.e.z.p.DockerSeleniumRemoteProxy - [21dc6e58d1713f05caa3cbb8cefc9d8d84a7895c845ac741fe1099b2be268ebb] is idle.
11:25:20.381 [AutoStartProxyPoolPoller] INFO  d.z.e.z.p.DockerSeleniumRemoteProxy - 21dc6e58d1713f05caa3cbb8cefc9d8d84a7895c845ac741fe1099b2be268ebb Shutting down node due to proxy being idle after test.
11:25:22.823 [pool-3-thread-1] INFO  o.o.g.s.proxy.DefaultRemoteProxy - 21dc6e58d1713f05caa3cbb8cefc9d8d84a7895c845ac741fe1099b2be268ebb Marking the node as down because it was stopped after 1 tests.
11:25:22.824 [pool-3-thread-1] WARN  d.z.e.z.registry.ZaleniumRegistry - Cleaning up stale test sessions on the unregistered node http://172.17.0.3:40000
11:25:22.889 [pool-3-thread-1] INFO  d.z.e.z.c.DockerContainerClient - Container 21dc6e58d1713f05caa3cbb8cefc9d8d84a7895c845ac741fe1099b2be268ebb does not exist - already shut down?.
```

The environment var that control this behaviour is `ZALENIUM_MAX_TEST_SESSIONS`, i.e. how many sessions can run on a node before it should be forcibly shutdown - defaults to `1`, but @pearj and I have been running at `20` without any issues. If you want to re-use containers, just increase this. Zalenium will still shutdown containers that have been idle if more than min / desired containers are running.

Thanks again for testing. Would you mind trying with `ZALENIUM_MAX_TEST_SESSIONS` set higher to make sure it works as expected on docker?

dennisgranath:
Ok maybe I just misunderstood the --desiredContainers flag? I interpreted as always keep this amount of containers up ready to serve incoming requests. When having ZALENIUM_MAX_TEST_SESSIONS set to 1 it will not remove the container when it until another request is coming in. It basically means that we are running the container even though it's quite useless.

dennisgranath:
I will try running with setting ZALENIUM_MAX_TEST_SESSIONS to higher value as well. Currently I run zalenium in minishift. I will se if I can deploy it to one of our openshift clusters instead...

dspasojevic:
@dennisgranath it *should* keep `desiredContainers` running. After the container reached max, it should have been removed after a few seconds and a new container built straight away. If it took until you ran another test, than that is a bug.

dennisgranath:
Thanks @dspasojevic then we share the common view! I think this will be great when we get everything to work:)

dspasojevic:
@dennisgranath I see why you are seeing this behaviour. At the moment, it is only checking for timed | invalid containers that should be shutdown when current > min. *But* it won't use a container that has reached the test limit to create a new session.

So once, you have reached the max sessions, the container will remain until another session comes in and causes a new container to start, bringing the count up to 3. Then the old, used up container will be shut down.

Like you say, this isn't good - the container won't be used, so it should be discarded straight away. I'll make that fix soon(ish).

pearj:
`ZALENIUM_MAX_TEST_SESSIONS` is the internal environment name. If you actually want to set that in the container, you need to use `MAX_TEST_SESSIONS`.

dspasojevic:
@dennisgranath / @pearj  if you have a chance, could you see if that latest version kills the containers that have reached max sessions immediately?

dennisgranath:
Sure thing @dspasojevic !

src/main/java/de/zalando/ep/zalenium/proxy/DockerSeleniumRemoteProxy.java
```diff
@@ -413,11 +413,15 @@ public boolean shutdownIfStale() {
             LOGGER.info(String.format("[%s] is stale.", getContainerId()));
             timeout("proxy being stuck | stale during a test.", ShutdownType.STALE);
             return true;
+        } else if (isTestSessionLimitReached() && !isBusy()) {
```
pearj: Why does it need to be on the stale one too? Does it run more often?

dspasojevic: Yep. Stale runs even when the pool doesn't have more than minimum desired number. If the check isn't also done here, containers that have run more than their max number of sessions won't be removed until after a new container is spun up to service another request.

Means that the effectively useless container is kept running longer than it should, and the test has to wait for a new container to be spun up (though this shouldn't take long).

It would also mean that if mix == max, then the pool would be full of useless containers and not be able to create a new one.

dennisgranath:
I think now it's working as expected @dspasojevic!

dspasojevic:
Thanks very much for testing @dennisgranath. I will wait for some review from @diemol then clean up tests and so on.

dspasojevic:
@dennisgranath good point. I disabled that health check in our environment a while ago - for the same reason - and forgot about it.

A zalenium specific end point would be good. @pearj is there something that we could usefully check?

pearj:
@dennisgranath just checking that `/wd/hub/status` returns a 200, is probably a good starting point, and in my opinion probably sufficient. Previously, it relied on the `DockerSeleniumStarterRemoteProxy` registering to the grid, which gave the illusion of capacity, but now that all that code is part of the registry directly, then checking for `"ready: true"` doesn't really have much meaning anymore.

We could probably create an API endpoint, but I'm not sure what we'd validate exactly, that we're able to communicate with Kubernetes/Docker daemon without error?

Once `/wd/hub/status` returns a 200 that means the selenium grid is online and will be able to start selenium nodes on demand when requested to do so.

dennisgranath:
Yes @pearj I saw that while checking the selenium code for healtcheck. We will remove status checking and only look for 200 for now.

src/main/java/de/zalando/ep/zalenium/proxy/AutoStartProxySet.java
```diff
@@ -0,0 +1,523 @@
...
+            LOGGER.debug(String.format("A node is coming up to handle this request."));
+            return true;
+        }
+
+        LOGGER.debug("No slots available, a new node will be created.");
+        return false;
+    }
+
+    private void dumpStatus() {
+        if (LOGGER.isInfoEnabled()) {
```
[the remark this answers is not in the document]
dennisgranath: or feature hide it somehow?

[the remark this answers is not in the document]
dspasojevic: Yep. Though might move to be a separate log category.

pearj:
:+1:

src/test/java/de/zalando/ep/zalenium/proxy/DockerSeleniumStarterRemoteProxyTest.java
```diff
@@ -1,459 +1,414 @@
 package de.zalando.ep.zalenium.proxy;
```
pearj: @dspasojevic Any reason why this isn't deleted altogther yet?

src/test/java/de/zalando/ep/zalenium/servlet/LiveNodeServletTest.java
```diff
@@ -1,131 +1,98 @@
...
-    
-    @After
-    public void tearDown() throws MalformedObjectNameException {
-        ObjectName objectName = new ObjectName("org.seleniumhq.grid:type=RemoteProxy,node=\"http://localhost:40000\"");
-        new JMXHelper().unregister(objectName);
-        objectName = new ObjectName("org.seleniumhq.grid:type=RemoteProxy,node=\"http://localhost:40001\"");
-        new JMXHelper().unregister(objectName);
-        ContainerFactory.setContainerClientGenerator(originalContainerClient);
-    }
+//    private GridRegistry registry;
```
pearj: @dspasojevic What's the plan for this test?

src/test/java/de/zalando/ep/zalenium/servlet/VncAuthenticationServletTest.java
```diff
@@ -1,172 +1,138 @@
...
-        
-        ArgumentCaptor<Integer> statusCaptor = ArgumentCaptor.forClass(Integer.class);
-        
-        vncAuthenticationServlet.doGet(request, response);
-        
-        verify(response).setStatus(statusCaptor.capture());
-        
-        assertThat(statusCaptor.getValue(), equalTo(403));
-    }
+//    private GridRegistry registry;
```
pearj: @dspasojevic and this one?

src/test/java/de/zalando/ep/zalenium/servlet/LiveNodeServletTest.java
dspasojevic: I will see whether I can rebirth this test.

src/test/java/de/zalando/ep/zalenium/servlet/VncAuthenticationServletTest.java
dspasojevic: Will check.

src/test/java/de/zalando/ep/zalenium/proxy/DockerSeleniumStarterRemoteProxyTest.java
dspasojevic: Was waiting to see if there was something that could be salvaged from here. I'll look at replacing with higher value integration tests.

src/test/java/de/zalando/ep/zalenium/servlet/LiveNodeServletTest.java
dspasojevic: I *think* that this test was written before the registry implementation could be swapped out. I'll see if I can rewrite to use a simpler grid impl so the docker container client doesn't have to be mocked.

Does that sound sensible to you @diemol ?

dspasojevic: Re-enabled.

src/test/java/de/zalando/ep/zalenium/servlet/VncAuthenticationServletTest.java
dspasojevic: Re-enabled.

pearj:
👍

src/main/java/de/zalando/ep/zalenium/proxy/AutoStartProxySet.java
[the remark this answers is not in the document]
dspasojevic: Added a separate category for the table. It is off by default in prod, on when running tests.

src/main/resources/logback.xml
```diff
@@ -17,6 +17,9 @@
     <!-- ALL TRACE DEBUG INFO WARN ERROR OFF -->
     <logger name="com.spotify.docker" level="WARN" />
     <logger name="org.seleniumhq.jetty9" level="WARN" />
+    
+    <!-- SET TO DEBUG TO GET STATUS LOGGING -->
+    <logger name="de.zalando.ep.zalenium.proxy.AutoStartProxySet.Status" level="INFO" />
```
dennisgranath: Do we need to set log level for this explicitly? The log level is set to logback.loglevel (defaulted to info) and can be set to debug using --debugEnabled flag to zalenium.

dspasojevic: It isn't necessary. In the past, I've explicitly (though unnecessarily) set debug levels to make it obvious to others how to turn on specific types of logging.

Happy to remove if that is the consensus :)

pearj:
👍

docker/Dockerfile
pearj: I've added a pull request to restore this bit with a build arg:
https://github.com/dspasojevic/zalenium/pull/3

src/main/resources/logback.xml
pearj: I wonder when this logging is most useful? For people new to zalenium? Which makes me think that it should be opt-out maybe? So default it to on, but have an environment variable that can turn it off for people who don't want it? Or maybe it's just too much noise to have it on by default?

I don't know 😉

peerster: When JSON logging is isn't enabled this should be on as default is my opinion

src/main/java/de/zalando/ep/zalenium/matcher/DockerSeleniumCapabilityMatcher.java
diemol: Yes and no, the idea of it is to reject as fast as possible a test request with a browser version we don't provide. 

We need to test what happens if I want to run a test with Chrome version 60, and we provide version 66. The request should be rejected since it should not match any nodes.

In the end, this is a minor thing since we can find ways to tweak it.

src/main/java/de/zalando/ep/zalenium/proxy/DockerSeleniumRemoteProxy.java
```diff
@@ -268,14 +315,15 @@ public void afterCommand(TestSession session, HttpServletRequest request, HttpSe
                 if (session.getSlot().getProxy() instanceof DockerSeleniumRemoteProxy) {
                     remoteName = ((DockerSeleniumRemoteProxy)session.getSlot().getProxy()).getRegistration().getContainerId();
                 }
-                ExternalSessionKey externalKey = Optional.ofNullable(session.getExternalKey()).orElse(new ExternalSessionKey("[No external key present]"));
                 LOGGER.info(String.format("Test session started with internal key %s and external key %s assigned to remote %s.",
                               session.getInternalKey(),
-                              externalKey.getKey(),
+                              session.getExternalKey().getKey(),
```
diemol: I had it like that initially, but sometimes the browser does not start for some reason and the whole thing crashes. In those cases `session.getExternalKey()` is null and the real exception is not sent back to the client.

As a matter of fact, I just bumped into it
```
22:04:16.497 [Terminate Test Session int id: [dccfea44-3436-4b26-ab35-dd9faf2311c6] ext id: [No external key was assigned] container: [f85d9cc697a6a4ff528d5930dbfed124fe856f9998295a03126831a60293c949]] INFO  d.z.e.z.p.DockerSeleniumRemoteProxy - f85d9cc697a6a4ff528d5930dbfed124fe856f9998295a03126831a60293c949 AFTER_SESSION command received. Node should shutdown soon...
22:04:16.500 [qtp1046665075-39] INFO  o.o.g.w.s.handler.RequestHandler - Error forwarding the new session null
java.lang.NullPointerException: null
	at de.zalando.ep.zalenium.proxy.DockerSeleniumRemoteProxy.afterCommand(DockerSeleniumRemoteProxy.java:320)
	at org.openqa.grid.internal.TestSession.forward(TestSession.java:281)
	at org.openqa.grid.web.servlet.handler.RequestHandler.forwardNewSessionRequestAndUpdateRegistry(RequestHandler.java:91)
	at org.openqa.grid.web.servlet.handler.RequestHandler.process(RequestHandler.java:114)
	at org.openqa.grid.web.servlet.DriverServlet.process(DriverServlet.java:86)
	at org.openqa.grid.web.servlet.DriverServlet.doPost(DriverServlet.java:70)
	at javax.servlet.http.HttpServlet.service(HttpServlet.java:707)
	at javax.servlet.http.HttpServlet.service(HttpServlet.java:790)
```

Could you please revert it?

diemol:
Thanks @dspasojevic for this PR!

Also thanks to @pearj, @dennisgranath and @peerster for the review and testing!

Overall, I really like what the PR is doing and goes in line with what we have discussed, I think it'll help a lot and solve other created issues.

We would really appreciate a couple of things:
* Please check the Codacy comments to see if something makes sense for you to fix.
* Would be cool to reenable most of the tests. I know that coverage doesn't guarantee working code always, but in the past those tests have saved us from screwing up a few things.

So, the proposal is the following:
* Please check my comments and let me know if it is ok to make those changes.
* Let's merge this to a protected branch, so we can run Travis with the Cloud providers as well.
* We review again how things look like, do some more testing and if more changes are needed, we can send a PR to that protected branch.

After merging and releasing this, I guess some users will create issues since we are changing slightly the behaviour of some parts (like the `desiredContainers` part). So it would be cool if you guys help us to answer those issues and also to update the docs (things that you already do often).

Let's target to release this in 1 or 2 weeks tops (hopefully earlier), what do you think?

src/main/java/de/zalando/ep/zalenium/proxy/AutoStartProxySet.java
```diff
@@ -0,0 +1,525 @@
+
+// Licensed to the Software Freedom Conservancy (SFC) under one
```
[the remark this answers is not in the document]
dspasojevic: Done.

src/main/java/de/zalando/ep/zalenium/proxy/DockerSeleniumRemoteProxy.java
```diff
@@ -147,7 +170,17 @@ public HtmlRenderer getHtmlRender() {
         Incrementing the number of tests that will be executed when the session is assigned.
      */
     @Override
-    public TestSession getNewSession(Map<String, Object> requestedCapability) {
+    public synchronized TestSession getNewSession(Map<String, Object> requestedCapability) {
```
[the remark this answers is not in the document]
dspasojevic: We need to ensure that the thread marking instances as timed out (which will start the container shutdown process) doesn't conflict with the thread assigning sessions to the proxy. We could make the locking finer grained - e.g. use a lock when checking and updating the timed out and last command flags, but the throughput here didn't seem to warrant it.

src/main/java/de/zalando/ep/zalenium/proxy/DockerSeleniumRemoteProxy.java
dspasojevic: Done.

dspasojevic:
Hi @diemol. Thanks for the feedback, I've actioned most of the changes and commented on the ones that I haven't. Codacy seems happy again.

There are two tests disabled at the moment. I'll see if I can fix them. I do wonder if more integration tests would be better - most of the unit tests that I've looked at seem to use a lot of mocking which (sometimes?) means that the tests are not as effective as they could be. wdyt?

src/main/java/de/zalando/ep/zalenium/proxy/DockerSeleniumRemoteProxy.java
```diff
@@ -369,25 +396,95 @@ public synchronized boolean isTestSessionLimitReached() {
...
+    	return this.timedOut.get();
     }
 
     /*
         Method to terminate an idle session via the registry, the code works because each one has only one slot
         We use BROWSER_TIMEOUT as a reason, but this could be changed in the future to show a more clear reason
      */
     @VisibleForTesting
-    protected synchronized void terminateIdleTest() {
+    protected void terminateIdleTest() {
```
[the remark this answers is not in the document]
pearj: This change removes that thread per proxy, there is a fixed threadpool of size 5 that handles terminations now:

See:
https://github.com/zalando/zalenium/pull/527/files#diff-cdb17f721a30ede405aeccaa422c3006R93

diemol:
Let's merge to a protected branch and we can take it from there, like that we can run Travis whole CI and get a better idea.

PR #533, opened 2018-04-20
Add initial Prometheus Support

pearj:
### Description
Add a Prometheus `/metrics` endpoint to allow monitoring of the zalenium server, initially the JVM and Jetty metrics.  In the future, we would add specific zalenium metrics.  This lays down the initial foundation.

To enable Prometheus to work, it needed to be integrated into the Jetty server, so AspectJ has been used to access the necessary classes.

Registration of the zalenium specific servlets has been moved into Aspect, so that we get greater control of the URLs that we can use.

I also fixed a bug where the cleanup servlet would crash if it tried to delete files/directories that didn't exist.

Additionally, the Nginx redirect problem has been fixed.

### Motivation and Context
This allows us to get greater insight into the performance of zalenium and how it uses resources.  Prometheus has become the defacto standard for monitoring kubernetes clusters, so it makes to most sense to use it.

### How Has This Been Tested?
I ran few multi-threaded selenium tests and checked the `/metrics` endpoint. I also tested the existing zalenium servlets to ensure they still operated.

### Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)

### Checklist:

- [x] My code follows the code style of this project.
- [x] I have read the **CONTRIBUTING** document.
- [x] All new and existing tests passed.

PR #534, opened 2018-04-24
Adding Prometheus Support

diemol:
Comes from #533

### DIFF
--- a/docker/nginx.conf
+++ b/docker/nginx.conf
@@ -12,10 +12,6 @@ http {
 
     server {
         listen 4444;
-
-        # Fix redirects
-        port_in_redirect off;
-        server_name_in_redirect off;
         
         satisfy any;
         allow 127.0.0.1;
@@ -34,10 +30,16 @@ http {
             proxy_set_header X-Real-IP $remote_addr;
         }
         location /dashboard {
+            # We want a relative redirect to work around reverse proxies infront of nginx
+            add_header Location "/dashboard/";
+            content_by_lua 'ngx.exit(301)';
+        }
+        location /dashboard/ {
             alias   /home/seluser/videos/;
             include /etc/nginx/mime.types;
             index   dashboard.html index.html;
         }
+
         location /dashboard/cleanup {
             proxy_pass http://127.0.0.1:4445/dashboard/cleanup;
             proxy_set_header Host      $host;
--- a/docker/nginx.conf
+++ b/docker/nginx.conf
@@ -13,6 +13,10 @@ http {
     server {
         listen 4444;
 
+        # Fix redirects
+        port_in_redirect off;
+        server_name_in_redirect off;
+        
         satisfy any;
         allow 127.0.0.1;
         auth_basic_user_file        /home/seluser/.htpasswd;
@@ -26,27 +30,21 @@ http {
 
         location / {
             proxy_pass http://127.0.0.1:4445;
+            proxy_set_header Host      $host;
+            proxy_set_header X-Real-IP $remote_addr;
         }
-        location /dashboard/ {
+        location /dashboard {
             alias   /home/seluser/videos/;
             include /etc/nginx/mime.types;
             index   dashboard.html index.html;
         }
         location /dashboard/cleanup {
-            proxy_pass http://127.0.0.1:4445/grid/admin/DashboardCleanupServlet;
+            proxy_pass http://127.0.0.1:4445/dashboard/cleanup;
             proxy_set_header Host      $host;
             proxy_set_header X-Real-IP $remote_addr;
         }
         location /dashboard/information {
-            proxy_pass http://127.0.0.1:4445/grid/admin/DashboardInformationServlet;
-            proxy_set_header Host      $host;
-            proxy_set_header X-Real-IP $remote_addr;
-        }
-        location /grid/console {
-            proxy_pass http://127.0.0.1:4445/grid/admin/ZaleniumConsoleServlet;
-        }
-        location /grid/admin/live {
-            proxy_pass http://127.0.0.1:4445/grid/admin/LivePreviewServlet;
+            proxy_pass http://127.0.0.1:4445/dashboard/information;
             proxy_set_header Host      $host;
             proxy_set_header X-Real-IP $remote_addr;
         }
@@ -71,7 +69,7 @@ http {
         }
         location /auth {
             internal;
-            proxy_pass http://127.0.0.1:4445/grid/admin/VncAuthenticationServlet;
+            proxy_pass http://127.0.0.1:4445/vnc/auth;
             proxy_pass_request_body off;
             proxy_set_header        Content-Length "";
             proxy_set_header        X-Original-URI $request_uri;
--- a/docker/nginx.conf
+++ b/docker/nginx.conf
@@ -26,8 +26,6 @@ http {
 
         location / {
             proxy_pass http://127.0.0.1:4445;
-            proxy_set_header Host      $host;
-            proxy_set_header X-Real-IP $remote_addr;
         }
         location /dashboard {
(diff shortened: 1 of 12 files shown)


diemol:
@pearj there were some CI errors, I'll double check locally and see if their are real.

pearj:
If this is creating problems with travis, maybe merge #527 onto a protected branch as well, and see how that one goes. There is more value with that change anyway.

pearj:
Wierd that it fails browserstack. It works on my pc™️ on browserstack.

pearj:
all the  logging is broken when running the tests since the move to logback I think:
```13:13:32,246 |-INFO in ch.qos.logback.classic.LoggerContext[default] - Could NOT find resource [logback-test.xml]
13:13:32,246 |-INFO in ch.qos.logback.classic.LoggerContext[default] - Could NOT find resource [logback.groovy]
13:13:32,246 |-INFO in ch.qos.logback.classic.LoggerContext[default] - Found resource [logback.xml] at [jar:file:/home/travis/build/zalando/zalenium/target/zalenium-3.11.0e-SNAPSHOT.jar!/logback.xml]
13:13:32,269 |-INFO in ch.qos.logback.core.joran.spi.ConfigurationWatchList@7dc0f706 - URL [jar:file:/home/travis/build/zalando/zalenium/target/zalenium-3.11.0e-SNAPSHOT.jar!/logback.xml] is not of type file
13:13:32,409 |-INFO in ch.qos.logback.classic.joran.action.ConfigurationAction - debug attribute not set
13:13:32,415 |-INFO in ch.qos.logback.classic.joran.action.LoggerContextListenerAction - Adding LoggerContextListener of type [ch.qos.logback.classic.jul.LevelChangePropagator] to the object stack
13:13:32,416 |-INFO in ch.qos.logback.classic.jul.LevelChangePropagator@4009e306 - Propagating DEBUG level on Logger[ROOT] onto the JUL framework
13:13:32,421 |-INFO in ch.qos.logback.classic.joran.action.LoggerContextListenerAction - Starting LoggerContextListener```
Running locally I had to do:
```-Dlogback.appender=STDOUT
-Dlogback.configurationFile=logback.xml```
So I guess that needs to be added into maven somehow when the tests run.
I added those jvm settings when running using the testng plugin inside eclipse

**Comment by diemol** [comment]
I'll check it

File: docker/nginx.conf [hunk 1]
```diff
@@ -26,27 +26,27 @@ http {
 
         location / {
             proxy_pass http://127.0.0.1:4445;
+            proxy_set_header Host      $host;
+            proxy_set_header X-Real-IP $remote_addr;
+        }
+        location /dashboard {
+            # We want a relative redirect to work around reverse proxies infront of nginx
+            add_header Location "/dashboard/";
+            content_by_lua 'ngx.exit(301)';
```
diemol: Somehow this did not work for me, I tried:

```
docker run --rm -ti --name zalenium -p 4450:4444 \
      -v /tmp/videos:/home/seluser/videos \
      -v /var/run/docker.sock:/var/run/docker.sock\
      dosel/zalenium start
```

So going to http://localhost:4450/dashboard should forward me and show the dashboard, but it did not.

diemol: Please omit the previous comment, I was using the wrong image 🤦‍♂️

diemol: It works!

diemol:
Somehow the BrowserStack test are failing, 
https://travis-ci.org/zalando/zalenium/jobs/371137450#L6358

I also reproduced it locally. I am not sure what the reason is, I guess when the response comes back from BrowserStack something in nginx doesn't get processed properly. 
Strange thing is that SauceLabs works fine.

docker/nginx.conf
```diff
@@ -26,27 +26,27 @@ http {
 
         location / {
             proxy_pass http://127.0.0.1:4445;
+            proxy_set_header Host      $host;
+            proxy_set_header X-Real-IP $remote_addr;
```
diemol: If I remove lines 29 and 30, BrowserStack works again.
The dashboard redirect also works.

Is there a reason to include them?

diemol:
Also, the little icons for cloud providers are not being rendered anymore. The console tries to find them at http://localhost:4444/grid/admin/ZaleniumResourceServlet/images/browserstack.png

image

diemol:
Nevermind, just found how to fix it and it works again.

image

docker/nginx.conf
diemol: OK, I'll remove them and push again.

src/main/java/de/zalando/ep/zalenium/aspect/HubAspect.aj
```diff
@@ -81,7 +81,7 @@ privileged public aspect HubAspect {
         handler.addServlet(LivePreviewServlet.class, "/grid/admin/live");
         handler.addServlet(ZaleniumConsoleServlet.class, "/grid/console");
         // We want resources to be at least 3 levels deep so that the prometheus servlet filter doesn't pick up all the individual resources
-        handler.addServlet(ZaleniumResourceServlet.class, "/grid/console/resources");
+        handler.addServlet(ZaleniumResourceServlet.class, "/grid/resources");
```
pearj: I think we should move the resources out of `/grid` so that it doesn’t appear in `/metrics` at all.

src/main/java/de/zalando/ep/zalenium/servlet/renderer/CloudProxyHtmlRenderer.java
```diff
@@ -87,13 +87,13 @@ private String getSingleSlotHtml(TestSlot s) {
         TestSession session = s.getSession();
         String icon = "";
         if (proxy instanceof TestingBotRemoteProxy) {
-            icon = "/grid/admin/ZaleniumResourceServlet/images/testingbot.png";
+            icon = "/grid/resources/images/testingbot.png";
```
pearj: Maybe just straight `/resources` pretty sure the hub doesn’t use it

pearj: Maybe let’s just merge and when we add zalenium Prometheus metrics. I’ll move the resources somewhere else now that I know what uses them.

pearj:
👍

src/test/java/de/zalando/ep/zalenium/it/ParallelIT.java
```diff
@@ -115,7 +115,7 @@ private WebDriver getWebDriver() {
     public void checkIframeLinksForLivePreview(String browserType, Platform platform) {
 
         // Go to the homepage
-        getWebDriver().get(String.format("http://%s:%s/grid/admin/live", ZALENIUM_HOST, ZALENIUM_PORT));
+        getWebDriver().get(String.format("http://%s:%s/grid/admin/live", hostIpAddress, ZALENIUM_PORT));
```
pearj: It could work in kubernetes in theory, if you run the test as a sidecar container (but it'd be fiddly), but just curious why we can't rely on `ZALENIUM_HOST` having the `hostIpAddress`? That way it can work for both docker and kubernetes.

What do we get out of finding the grid by using the current IP address inside the test? All it seems to do is ensure that maven is running on the same IP address as zalenium?

peerster:
👍

src/test/java/de/zalando/ep/zalenium/it/ParallelIT.java
diemol: The purpose of it was to check that the live view was reachable also by its ip and that the headers were being forwarded properly, since we broke that once. 

When the live view is done again, we can just do it from scratch.

diemol:
👍

diemol:
Thanks @pearj!

PR #541, opened 2018-04-25
Dummy commit to trigger CI

diemol:
Just to check if the BrowserStack errors in #533 are related to the Nginx changes or to BrowserStack itself


diemol:
OK, CI passes, so it should be related to the nginx changes,

Issue #518, opened 2018-04-06
Tests fail because OpenShift container vanishes underneath


Lehtux:
Zalenium Image Version(s):  
3.11.0d
Selenium image:
3.11.0-p5

OS: 
OpenShift Enterprise

## Expected Behavior -
Tests will pass through normally.
## Actual Behavior -
Seems like Zalenium is closing containers even though a test case is registered there?
Getting this error from hub under "heavy load":
`15:10:11.675 [qtp565372776-2332] INFO  o.o.g.w.s.handler.RequestHandler - Error forwarding the new session Error forwarding the request unexpected end of stream on Connection{10.1.16.158:40049, proxy=DIRECT hostAddress=/10.1.16.158:40049 cipherSuite=none protocol=http/1.1}
org.openqa.grid.internal.exception.NewSessionException: Error forwarding the request unexpected end of stream on Connection{10.1.16.158:40049, proxy=DIRECT hostAddress=/10.1.16.158:40049 cipherSuite=none protocol=http/1.1}
	at org.openqa.grid.web.servlet.handler.RequestHandler.forwardNewSessionRequestAndUpdateRegistry(RequestHandler.java:94)
	at org.openqa.grid.web.servlet.handler.RequestHandler.process(RequestHandler.java:114)
	at org.openqa.grid.web.servlet.DriverServlet.process(DriverServlet.java:86)
	at org.openqa.grid.web.servlet.DriverServlet.doPost(DriverServlet.java:70)
	at javax.servlet.http.HttpServlet.service(HttpServlet.java:707)
	at javax.servlet.http.HttpServlet.service(HttpServlet.java:790)
	at org.seleniumhq.jetty9.servlet.ServletHolder.handle(ServletHolder.java:860)
	at org.seleniumhq.jetty9.servlet.ServletHandler.doHandle(ServletHandler.java:535)
	at org.seleniumhq.jetty9.server.handler.ScopedHandler.nextHandle(ScopedHandler.java:188)
	at org.seleniumhq.jetty9.server.session.SessionHandler.doHandle(SessionHandler.java:1595)
	at org.seleniumhq.jetty9.server.handler.ScopedHandler.nextHandle(ScopedHandler.java:188)
	at org.seleniumhq.jetty9.server.handler.ContextHandler.doHandle(ContextHandler.java:1253)
	at org.seleniumhq.jetty9.server.handler.ScopedHandler.nextScope(ScopedHandler.java:168)
[...]

PR #548, opened 2018-04-30
Teardown re-registered proxies so that their threads are stopped gc'd.

dspasojevic:
Fixes a race condition where a proxy re-registers after it has been asked to stop and then continues to poll itself and attempt to shutdown.

### DIFF
--- a/src/main/java/de/zalando/ep/zalenium/proxy/AutoStartProxySet.java
+++ b/src/main/java/de/zalando/ep/zalenium/proxy/AutoStartProxySet.java
@@ -166,11 +166,19 @@ public class AutoStartProxySet extends ProxySet implements Iterable<RemoteProxy>
     }
 
     public void add(RemoteProxy proxy) {
+        boolean shouldAdd = true;
         if (proxy instanceof DockerSeleniumRemoteProxy) {
             DockerSeleniumRemoteProxy dockerSeleniumRemoteProxy = (DockerSeleniumRemoteProxy) proxy;
-            this.register(dockerSeleniumRemoteProxy);
+            shouldAdd = this.register(dockerSeleniumRemoteProxy);
         }
-        super.add(proxy);
+
+        if (shouldAdd) {
+            super.add(proxy);
+        }
+        else {
+            // Won't be tracking the proxy, so it won't be removed and shutdown later - tear down.
+            proxy.teardown();
+        }
     }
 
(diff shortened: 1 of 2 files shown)


peerster:
Is this a fix for #518 ?

dspasojevic:
Kind of. The larger PR on this branch already fixed the symptom of the test failing. This PR gets rid of annoying log messages and a useless timer thread.

If you’ve tried the main PR and still experienced failed tests because of this issue, I would be interested to hear 😀

peerster:
@dspasojevic  Ah ok. No, I haven't seen this problem when we tested the main PR. I was just a bit curious because of the title suggested that this solves that problem, even though as you said the main PR already seem to have fixed it 😄

diemol:
Thanks @dspasojevic! 
Let me test more #527 and fix the failing job in Travis tomorrow May 1st. I think this should fix that issue with proxies registering again, but I prefer to fix that on the Proxy class directly, so the request is not sent anymore. 

I'll let you know soon.

dspasojevic:
Hi @diemol - just to clarify that I think the cause of this is the 'normal' behaviour of the selenium nodes.

Because the registration request is handled by a different thread to that which is monitoring containers to be killed, it is possible for a proxy to be killed and removed from the proxy set at the same time that the registration request is being processed.

The remote node is doing the 'right' thing here I think.

When you say you will fix it in the proxy, do you mean fix it in the remote node? Or change the DockerSeleniumRemoteProxy to signal to the remote node that it is being shutdown and should stop trying to register?

pearj:
@diemol I think the registration request happens on a schedule from the selenium node as the polling interval or registration interval I think?

diemol:
@dspasojevic @pearj you guys are right, I always mix it up in this topic.

What each node has is a health check that unregisters the node after X attempts. This is the code we can control on our side.

The registering part, is running in the JVM inside the docker-selenium containers, here is the code for that part https://github.com/SeleniumHQ/selenium/blob/selenium-3.11.0/java/server/src/org/openqa/grid/internal/utils/SelfRegisteringRemote.java#L168, and we cannot control it obviously. 

Before, we were starting the nodes with `registerCycleInterval=0`, but then sometimes nodes were not registering in Kubernetes and they ended up as zombies. Now we have some random value for retrying in those cases.

So I think this PR makes a lot of sense, thanks!

PR #550, opened 2018-05-02
Fix unit tests

dspasojevic:
Fixes all unit tests.

Changed usages of `ZaleniumRegistry` to `SimpleRegistry` when testing unrelated classes.


pearj:
👍

PR #555, opened 2018-05-05
Fixes issue 552

diemol:
When requests can be served by docker-selenium but are sent to cloud providers because no docker-selenium node is running.


elgalu:
:+1:

diemol:
👍

PR #556, opened 2018-05-06
Allows changing the ubuntu mirror via a build argument

pearj:
* Also allows removing of dependencies not needed in Kubernetes


pearj:
👍

PR #557, opened 2018-05-06
Change Ubuntu Mirror via Args

diemol:


diemol:
👍

peerster:
👍

peerster:
@pearj How much smaller does the kubernetes version get?

Wondering if maybe it could be worth the trouble of making two docker builds; standard and kubernetes version.

pearj:
@peerster It is only 37 MB less.  It's more that it speeds up build time over here in Australia, so we don't have to download unneeded bits and pieces.

pearj:
I guess you might gain more size decreases if you strip out the Spotify docker library inside the zalenium jar itself. But for that to work, we'd need to make zalenium more modular, so that the kuberentes and docker pieces could be swapped in and out, without compile issues. 

But I don't think we'd gain very much for doing that.

PR #574, opened 2018-05-14
Upgrade to Selenium 3.12.0

diemol:


src/main/java/de/zalando/ep/zalenium/proxy/DockeredSeleniumStarter.java
```diff
@@ -349,4 +341,34 @@ private TimeZone getConfiguredTimeZoneFromCapabilities(Map<String, Object> reque
 
         return timeZone;
     }
+
```
pearj: @diemol you might need to update the dockerclient too as it searches by IP address to find containers on registration. Maybe that approach won’t work in —net=host mode? Unless of course it just works anyway.

diemol: I did it wrong the first time and then the docker container client did not find them at all. Afterwards I saw the error and made the changes to pass the updated port as a parameter to it. It worked fine after that :)

diemol:
👍

Issue #491, opened 2018-03-20
Zalenium versions that use Selenium 3.11 intermittently fail re: "No System TLS"


borisivan:
Zalenium Image Version(s):  
All versions that use Selenium 3.11

Docker Version:  
17.09, others

Docker Command to start Zalenium:  
docker, run, -tid, --name=obfuscated, --net=host, -p, 4444:4444, -v, /var/run/docker.sock:/var/run/docker.sock, -v, /usr/bin/docker:/usr/bin/docker, -v, /dev/shm:/dev/shm, -v, /tmp/videos/obfuscated:/home/seluser/videos, -v, /tmp/Downloads/obfuscated:/tmp/node/home/seluser/Downloads, --privileged, dosel/zalenium:latest, start, --chromeContainers, 7, --screenWidth, 1920, --seleniumImageName, elgalu/selenium:latest, --screenHeight, 1080, --firefoxContainers, 0, --keepOnlyFailedTests, true, --videoRecordingEnabled, true, --debugEnabled, false, --maxTestSessions, 150

## Expected Behavior -
With 6 tests running in parallel across those 7 static containers, tests run normally. (same thing happens w/ 6 containers, I have 7 to help re: availability during cleanup)

## Actual Behavior -
Out of ~240 tests, I will have 1-5 fail with the following footprint:

```
org.openqa.selenium.WebDriverException: cannot forward the request No System TLS
Command duration or timeout: 0 milliseconds
	at sun.reflect.NativeConstructorAccessorImpl.newInstance0(Native Method)
	at sun.reflect.NativeConstructorAccessorImpl.newInstance(NativeConstructorAccessorImpl.java:62)
	at sun.reflect.DelegatingConstructorAccessorImpl.newInstance(DelegatingConstructorAccessorImpl.java:45)
	at java.lang.reflect.Constructor.newInstance(Constructor.java:423)
	at org.openqa.selenium.remote.ErrorHandler.createThrowable(ErrorHandler.java:214)
	at org.openqa.selenium.remote.ErrorHandler.throwIfResponseFailed(ErrorHandler.java:166)
	at org.openqa.selenium.remote.http.JsonHttpResponseCodec.reconstructValue(JsonHttpResponseCodec.java:40)
	at org.openqa.selenium.remote.http.AbstractHttpResponseCodec.decode(AbstractHttpResponseCodec.java:80)
[...]

PR #597, opened 2018-05-27
Adding retry to getHttpClient when "No System TLS" pops up #491

diemol:
To solve #491

### DIFF
--- a/src/main/java/de/zalando/ep/zalenium/registry/ZaleniumRegistry.java
+++ b/src/main/java/de/zalando/ep/zalenium/registry/ZaleniumRegistry.java
@@ -30,6 +30,7 @@ import java.time.Clock;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
+import java.util.Random;
 import java.util.Set;
 import java.util.concurrent.CopyOnWriteArrayList;
 import java.util.concurrent.TimeUnit;
@@ -454,26 +455,29 @@ public class ZaleniumRegistry extends BaseGridRegistry implements GridRegistry {
     public HttpClient getHttpClient(URL url) {
         // https://github.com/zalando/zalenium/issues/491
         int maxTries = 3;
-        for (int i = 0; i < maxTries; i++) {
+        for (int i = 1; i <= maxTries; i++) {
             try {
                 HttpClient client = httpClientFactory.createClient(url);
-                if (i > 0) {
-                    String message = String.format("Successfully created HttpClient for url %s, after attempt #%s", url, (i+1));
+                if (i > 1) {
+                    String message = String.format("Successfully created HttpClient for url %s, after attempt #%s", url, i);
                     LOG.warn(message);
                 }
                 return client;
             } catch (Exception | AssertionError e) {
-                String message = String.format("Error while getting the HttpClient for url %s, attempt #%s", url, (i+1));
+                String message = String.format("Error while getting the HttpClient for url %s, attempt #%s", url, i);
                 LOG.debug(message, e);
-                if (i == 2) {
+                if (i == maxTries) {
                     throw e;
+                }
+                try {
+                    Thread.sleep((new Random().nextInt(5) + 1) * 1000);
+                } catch (InterruptedException exception) {
+                    LOG.error("Something went wrong while delaying the HttpClient creation after a failed atttempt", exception);
                 }
             }
         }
-        LOG.warn(String.format("Last attempt to get the HttpClient for url %s", url));
-        return httpClientFactory.createClient(url);
+        throw new IllegalStateException(String.format("Something went wrong while creating a HttpClient for url %s", url));
     }
-
 
     /**
      * iterates the queue of incoming new session request and assign them to proxy after they've been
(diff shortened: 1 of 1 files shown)


diemol:
👍

src/main/java/de/zalando/ep/zalenium/registry/ZaleniumRegistry.java
```diff
@@ -449,6 +450,31 @@ public void uncaughtException(Thread t, Throwable e) {
...
+                return client;
+            } catch (Exception | AssertionError e) {
+                String message = String.format("Error while getting the HttpClient for url %s, attempt #%s", url, (i+1));
+                LOG.debug(message, e);
+                if (i == 2) {
+                    throw e;
+                }
+            }
+        }
+        LOG.warn(String.format("Last attempt to get the HttpClient for url %s", url));
```
pearj: Will this log every time a client is requested? It looks this way?

src/main/java/de/zalando/ep/zalenium/registry/ZaleniumRegistry.java
```diff
@@ -449,6 +450,31 @@ public void uncaughtException(Thread t, Throwable e) {
...
+                    String message = String.format("Successfully created HttpClient for url %s, after attempt #%s", url, (i+1));
+                    LOG.warn(message);
+                }
+                return client;
+            } catch (Exception | AssertionError e) {
+                String message = String.format("Error while getting the HttpClient for url %s, attempt #%s", url, (i+1));
+                LOG.debug(message, e);
+                if (i == 2) {
+                    throw e;
+                }
```
pearj: So, you don't think there should be a random sleep? Just retrying it straight away seems to fix it in all cases?

src/main/java/de/zalando/ep/zalenium/registry/ZaleniumRegistry.java
```diff
@@ -449,6 +450,31 @@ public void uncaughtException(Thread t, Throwable e) {
...
+                HttpClient client = httpClientFactory.createClient(url);
+                if (i > 0) {
+                    String message = String.format("Successfully created HttpClient for url %s, after attempt #%s", url, (i+1));
+                    LOG.warn(message);
+                }
+                return client;
+            } catch (Exception | AssertionError e) {
+                String message = String.format("Error while getting the HttpClient for url %s, attempt #%s", url, (i+1));
+                LOG.debug(message, e);
+                if (i == 2) {
```
pearj: This should have some relation to `maxTries`. If you started with `i=1` then you could use `i == maxTries` instead

src/main/java/de/zalando/ep/zalenium/registry/ZaleniumRegistry.java
```diff
@@ -449,6 +450,31 @@ public void uncaughtException(Thread t, Throwable e) {
         }
     }
 
+    @Override
+    public HttpClient getHttpClient(URL url) {
+        // https://github.com/zalando/zalenium/issues/491
+        int maxTries = 3;
+        for (int i = 0; i < maxTries; i++) {
```
pearj: Maybe `for (int i = 1; i <= maxTries; i++)` might work better here, so you can handle the exception throw easier.

src/main/java/de/zalando/ep/zalenium/registry/ZaleniumRegistry.java
pearj: Oh my bad. There is a return statement there. I read it too quickly.

src/main/java/de/zalando/ep/zalenium/registry/ZaleniumRegistry.java
```diff
@@ -449,6 +450,31 @@ public void uncaughtException(Thread t, Throwable e) {
...
+            } catch (Exception | AssertionError e) {
+                String message = String.format("Error while getting the HttpClient for url %s, attempt #%s", url, (i+1));
+                LOG.debug(message, e);
+                if (i == 2) {
+                    throw e;
+                }
+            }
+        }
+        LOG.warn(String.format("Last attempt to get the HttpClient for url %s", url));
+        return httpClientFactory.createClient(url);
```
pearj: Do we actually need this return at all? I think it might be dead code. Because when it fails it throws an exception.

src/main/java/de/zalando/ep/zalenium/registry/ZaleniumRegistry.java
pearj: I’m not sure if that will log at all will it? Since it’s after the throw and outside the for loop

diemol: Yes, that is dead code. I'll remove it, I need to leave the return though.

src/main/java/de/zalando/ep/zalenium/registry/ZaleniumRegistry.java
diemol: Right that is more simple.

src/main/java/de/zalando/ep/zalenium/registry/ZaleniumRegistry.java
diemol: 👍

src/main/java/de/zalando/ep/zalenium/registry/ZaleniumRegistry.java
diemol: It worked right away in all cases I tried. But I agree with you, it is a good idea to add a small delay, random between 1 to 5 seconds.

diemol:
👍

elgalu:
:+1:

PR #608, opened 2018-06-05
Fix bug with desired containers counting as waiting test sessions

pearj:
### Description
Fix bug with desired containers counting as waiting test sessions

* Also make it possible to override bucket size for the test start latency metric.
* Change SessionRequestFilter to use get instead of looping through the map

### How Has This Been Tested?
Manually tested on an OpenShift Cluster, starting concurrent tests with a desiredCapabilities of 2, and ensuring the metric goes back to 0 after no tests are running.

Additionally I changed the bucket size by setting an environment variable of:
`ZALENIUM_TEST_SESSION_LATENCY_BUCKETS=0.5,2.5,5,10,11,12,13,14,15,20,30,60`

I also set `NEW_SESSION_WAIT_TIMEOUT=5000` so that most requests would timeout when waiting for a new container to start, and verified that `selenium_test_sessions_waiting` was decremented for timed out sessions.

### Types of changes

### Checklist:

### DIFF
--- a/src/main/java/de/zalando/ep/zalenium/registry/ZaleniumRegistry.java
+++ b/src/main/java/de/zalando/ep/zalenium/registry/ZaleniumRegistry.java
@@ -24,7 +24,11 @@ import org.openqa.selenium.remote.server.log.LoggingManager;
 import de.zalando.ep.zalenium.proxy.AutoStartProxySet;
 import de.zalando.ep.zalenium.proxy.DockerSeleniumRemoteProxy;
 import de.zalando.ep.zalenium.proxy.DockeredSeleniumStarter;
+import de.zalando.ep.zalenium.util.Environment;
 import de.zalando.ep.zalenium.util.ZaleniumConfiguration;
+import io.prometheus.client.Collector;
+import io.prometheus.client.Gauge;
+import io.prometheus.client.Histogram;
 
 import java.net.URL;
 import java.time.Clock;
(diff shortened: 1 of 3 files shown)


elgalu:
:+1:

pearj:
:+1:

pearj:
I've done a bit more testing and I've discovered that sometimes active sessions is also mis-counted. I might change the approach to read the information directly from `de.zalando.ep.zalenium.registry.ZaleniumRegistry.activeTestSessions` on-demand, instead of trying to figure out the right places to increment and decrement.  That way I could add labels per proxy type, so that you could see all the different cloud provider active sessions vs the docker ones.

diemol:
Where is this variable set? `NEW_SESSION_WAIT_TIMEOUT`, I don't see it in the changes

src/main/java/de/zalando/ep/zalenium/registry/ZaleniumRegistry.java
```diff
@@ -58,6 +62,23 @@
...
     private final List<RemoteProxy> registeringProxies = new CopyOnWriteArrayList<>();
     private volatile boolean stop = false;
+    
+    private static final Environment defaultEnvironment = new Environment();
+    
+    private static final double[] DEFAULT_TEST_SESSION_LATENCY_BUCKETS =
+            new double[] { 0.5,2.5,5,10,15,20,25,30,35,40,50,60 };
+
+    // Allows overriding of the test session latency buckets for prometheus
+    private static final String ZALENIUM_TEST_SESSION_LATENCY_BUCKETS = "ZALENIUM_TEST_SESSION_LATENCY_BUCKETS";
```
pearj: Here. I didn’t bother to plumb it through zalenium.sh

pearj: Adding it to zalenium.sh is only needed if you want to use command line args. Can add that if you like. But that’s more of a docker thing. 

Maybe we should have a table in the documentation that shows mapping’s between environment variables and command line args where they exist.

diemol: OK, so I think we were talking about different variables. Since in the initial comment you mention [`NEW_SESSION_WAIT_TIMEOUT`](https://github.com/zalando/zalenium/blob/9abf3a2d26d373e4be6611802cc6dad01489ce5d/scripts/zalenium.sh#L25), I thought that one was changing. 

But I guess you were only talking about `ZALENIUM_TEST_SESSION_LATENCY_BUCKETS`, all good :)

pearj: Oh yeah, sorry. I mentioned the `NEW_SESSION_WAIT_TIMEOUT` as that is how I tested test session requests being timed out, and the metrics still decrementing properly.

Issue #430, opened 2018-02-07
Feature request: Send video and logs to some external system for storage, indexing and reporting


adrichem:
Zalenium Image Version(s):  3.8.1j
Docker Version:  N/A
docker-compose, version: N/A
OS: N/A
Docker Command to start Zalenium:  N/A

## Expected Behavior -
Our tests produce artifacts that we want  for later analysis. We use things like blobstorage, Elasticsearch and Kibana to store and index them along with relevant metadata. The video recordings that Zalenium creates are currently difficult to deal with because:
- Scraping the file-system will be error-prone as the files can be overwritten by the dashboard code at any time
- I prefer to avoid persistent volumes for containers.
- The metadata in the dashboard html is interesting, however we want to decide at run-time what extra metadata is associated with each artifact.

On https://github.com/adrichem/zalenium ( branch artifact-store), I have created a beta version that can:

1. Send video and logs to some external system for storage and indexing. Its a simple HTTP Post. The dashboard is not affected and remains in place.
1. Include user supplied metadata into that Post request. We use the cookie mechanism just like for the test result

Before I open a PR, I was wondering how you feel about such a change? And if you can point out any work that needs to be done to make it good enough for a PR.

I hope you find it useful,
Regards

## Actual Behavior -
Currently Zalenium has its own dashboard that testers user to look at videorecordings and logfiles. It does not yet allow these artifacts to be stored and indexed by some external system.

diemol:
Hi @adrichem,

This is quite a surprise to be honest, and I kind of like the idea. But my main concern is that the Zalenium container is already doing a lot of stuff (dashboard logic, copying files locally, nginx proxy) in addition to handle test requests, and adding more logic would make things more complicated.

We are starting to work slowly into redoing the dashboard, to remove some of the heavy work that it is done by updating it directly in the core logic where test requests are handled. The solution is nothing really fancy, but just exposing the data the dashboard uses through a JSON in an endpoint, and then the dashboard reads this every X seconds.

What do you think if we do this as a separate service running in a side container, where it could read that endpoint and then just pull the data from the Zalenium container and then put it wherever is needed?

PS: Thanks a lot for your interest and intention to improve Zalenium.

adrichem:
Hi @diemol,

I have a few reasons why I think its best to use the fire-and-forget HTTP Post approach as coded in the beta version:

[...]

PR #613, opened 2018-06-07
Sends video and logs to some external system for storage, indexing and reporting. 

adrichem:
### Description
As discussed in https://github.com/zalando/zalenium/issues/430. I'm not sure how your documentation works. So I didn't update that. I hope you will do that. It needs to show, that
1. Setting an environment variable tells Zalenium where to send the artifacts to the store. My compose file uses this:
``` - REMOTE_DASHBOARD_HOST=http://artifact-store:4000 ```
1. If that environment variable does not exist, then the local dashboard will be used and Zalenium will not send the artifacts to the store.
1. The tester can tell Zalenium to send arbitrary JSON metadata together with the files. Here's a snippet of C# code that does that:
```
this.Driver.Manage().Cookies.AddCookie(new Cookie("zaleniumMetadata", JsonConvert.SerializeObject(Metadata)))
```

### Motivation and Context
See https://github.com/zalando/zalenium/issues/430. 

### How Has This Been Tested?
Been running it in our test environment for a few months. Today's commit implements an XOR choice between local or remote dashboard. That was tested by
1.  starting it with only local dashboard, running a session and seeing the video appear on the local dashboard.
2. starting it with remote dashboard, running a session and seeing that the dashboard remains empty.

### Types of changes

### Checklist:
- [ ] I have read the **CONTRIBUTING** document.

### CODE REVIEW

**Comment by diemol** [comment]
Thanks for this @adrichem!

Is it possible for you to rebase first? I know you mentioned in the issue that you were short of time but if you can, that would be really great. Afterwards, based on your instructions, I can update the docs.

**Comment by adrichem** [comment]
You're welcome of course. 

I've rebased my branch onto your master. Can you please take a good look at the final set of diffs? I had a lot of conflicts due to age of the branch and the rename /replace of Dashboard.java to LocalDashboard.java .

Also please let me know if I need to tackle the Codacy\PR Quality Review issues,.

Cheers!

File: src/main/java/de/zalando/ep/zalenium/dashboard/remote/RemoteDriverLogDashboard.java [hunk 1]
```diff
@@ -0,0 +1,44 @@
...
+    public void updateDashboard(TestInformation testInformation) throws Exception {
+        if (null == testInformation) throw new IllegalArgumentException("testInformation");
+
+        if(!isEnabled()) {
+            return;
+        }
+
+        List<FormField> fields = new ArrayList<>();
+
+        fields.add(new FormFile() {{
```
pearj: I don't think we want the "double-brace initialisation" anywhere, as it will probably create a garbage collection overhead, as well as potentially leaking memory.

Since this called after every test, it has more chance of being a problem.

I didn't even know that this "double-brace initialisation" was even a thing, until codacy picked it up: https://app.codacy.com/app/diemol_zalenium/zalenium/pullRequest?bid=4117327&prid=1752756#newIssuesView

I did a little digging and came across this article explaining why it's not a great idea: https://blog.jooq.org/2014/12/08/dont-be-clever-the-double-curly-braces-anti-pattern/

src/test/java/de/zalando/ep/zalenium/dashboard/LocalDashboardTest.java
```diff
@@ -36,54 +36,76 @@
...
-        Dashboard.updateDashboard(ti);
-        Assert.assertEquals(2, Dashboard.getExecutedTests());
-        Assert.assertEquals(2, Dashboard.getExecutedTestsWithVideo());
+        LocalDashboard dashboard = new LocalDashboard();
+        try {
+            dashboard.updateDashboard(ti);
+            dashboard.updateDashboard(ti);
+            Assert.assertEquals(2, LocalDashboard.getExecutedTests());
+            Assert.assertEquals(2, LocalDashboard.getExecutedTestsWithVideo());
+        } catch (Exception e) {
```
pearj: Why do you catch the exception and rethrow it? I don't think you need to catch the exception at all. All it means is that the test errors will be harder to find the actual cause because the original exception will be lost.

adrichem:
Alright. I fixed almost all the Codacy issues and the unnecessary try/catch on the tests for the local dashboad (Doh!) One issue I did not fix was the warning about one unittest not containing an assert() or fail(). That's because the  .thenThrow() causes the assert() behavior. I'm not familiar enough with mockito to see how I could improve that.

 I also:
1. Reverted the rename and replace of Dashboard.java / LocalDashboard.java. It was  way to hard to check that I didn't accidentally lose any changes w.r.t. your master. Squashed those commits  i my branch so that we don't keep running into that again and again
1. Fixed a few compiler warnings introduced in the PR
1. Hit the 'Update branch' button in the PR to merge your latest commits from master

 Should  be good to go  now. If not, feedback appreciated!

diemol:
Thanks @adrichem for fixing the Codacy issues, I will merge this to a protected branch so Travis can run all the tests. After that, add some docs based on your comments and release. Hopefully it won't take so long.

Idsbali:
Can we also send other files (files getting downloaded in the browser) to some external system storage (like S3) ?

PR #630, opened 2018-06-27
Run an E2E test in CDP.

bmaher:
### Description
Introduce an end-to-end test to Zalenium, running in Zalando's CI system.

### Motivation and Context
The test consists of:
- A basic Express application
- A basic WDIO test
- A docker-compose consisting of an instance of Zalenium, the application, and the test runner

The idea is that this represents the most basic use-case for Zalenium, and therefore, no change should break this setup.

### How Has This Been Tested?
This has been tested locally using docker-compose to run the test itself.

### Types of changes

### Checklist:

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,5 @@ logs/
 TODO.private.md
 velocity.log
 docs/_site
+node_modules
+errorShots
--- a/delivery.yaml
+++ b/delivery.yaml
@@ -1,0 +1,42 @@
+version: "2017-09-20"
+
+pipeline:
+- id: build
+  type: script
+  commands:
+  - desc: Install dependencies
+    cmd: |
+      # Install Docker
+      curl -fsSL https://delivery.cloud.zalando.com/utils/ensure-docker | sh
+      # Install docker-compose
+      curl -L https://github.com/docker/compose/releases/download/1.21.2/docker-compose-`uname -s`-`uname -m` > docker-compose
+      chmod +x docker-compose
+      sudo mv docker-compose /usr/local/bin
+      # Install Maven and JDK
+      apt-get update
+      apt-get install -y maven openjdk-8-jdk-headless
+      # Install docker-selenium
+      docker pull elgalu/selenium
+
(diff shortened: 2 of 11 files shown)


diemol:
Thanks @bmaher! Good idea.

Only one thing, what do you think about using this script (or an adaptation) instead of `wait-for-it.sh`?

```
#!/bin/bash
# wait-for-grid.sh

set -e

cmd="$@"

while ! curl -sSL "http://hub:4444/wd/hub/status" 2>&1 \
        | jq -r '.value.ready' 2>&1 | grep "true" >/dev/null; do
    echo 'Waiting for the Grid'
    sleep 1
done

>&2 echo "Selenium Grid is up - executing test"
exec $cmd
``` 

I suggest this since having the port 4444 up and listening is no guarantee that the grid is still able to receive test requests (it might be possible that no nodes have registered yet).

bmaher:
@diemol Sure thing, looks good; I'll switch it out.

diemol:
@bmaher everything looks good to me.

The only question I have is: what do we do when the CDP build fails? I don't have access anymore to the logs and most of the coding nowadays is done during weekends and evenings, so I doubt someone can help right away.

Do you guys have something in mind for those cases?

bmaher:
@diemol Yes I was thinking about that too, I have two ideas:

1. If the build status reports a failure, and running the tests locally show no failure, i.e. a potential CDP issue, then we'd wait until working hours to find a resolution
2. We could make the CDP status non-blocking for a merge, mark the images in some way, and re-verify during working hours

diemol:
I think option 1 sounds fine.

About option 2, I don't think we should mark the CDP build as required to merge. The logs are not accessible publicly and this would go against the Open Source nature of the project, all related things to it must be open and accessible by anyone who wants to contribute.

Besides that, looks great. Thanks for adding those. Extra points would be if we run the same tests in Travis, what do you think @bmaher ?

diemol:
👍

elgalu:
👍

PR #635, opened 2018-07-03
Change ownership before coping dashboard files

dusansusic:
This will fix permissions issues when coping dashboard files on volume which is automatically created and mounted by k8s.

### DIFF
--- a/scripts/zalenium.sh
+++ b/scripts/zalenium.sh
@@ -399,6 +399,7 @@ StartUp()
     fi
 
     echo "Copying files for Dashboard..."
+    sudo chown -R seluser.seluser  /home/seluser/videos
     cp /home/seluser/index.html /home/seluser/videos/index.html
     cp -r /home/seluser/css /home/seluser/videos
     cp -r /home/seluser/js /home/seluser/videos


scripts/zalenium.sh
```diff
@@ -399,6 +399,7 @@ StartUp()
     fi
 
     echo "Copying files for Dashboard..."
+    sudo chown -R seluser.seluser  /home/seluser/videos
```
pearj: I've never seen the dot syntax for chown before? Pretty sure it should be `seluser:seluser`, the man page says that too.
Also, this isn't safe to merge, because sudo is not always available, especially in OpenShift:
```
$ sudo chown -R seluser.seluser  /home/seluser/videos
sudo: PERM_SUDOERS: setresuid(-1, 1, -1): Operation not permitted
sudo: no valid sudoers sources found, quitting
sudo: setresuid() [0, 0, 0] -> [1000090000, -1, -1]: Operation not permitted
sudo: unable to initialize policy plugin
```

You'll need to do something like this:
```
if [ "${WE_HAVE_SUDO_ACCESS}" == "true" ]; then
    sudo chown -R seluser:seluser /home/seluser/videos
fi
```

pearj: Thinking about it more, this more sounds like a fault with your kubernetes cluster.  When you create a persistent volume kuberentes should ensure the permissions are correct.  You shouldn't need sudo and changing permissions.

pearj:
Since you appear to be the only person with this problem I think a better approach would be for you to use an init-container to fix the permissions problem. As all containers in a pod share the same volume mounts.  I don't think this change is necessary, for people that have sudo access enabled it will just slow the startup time

PR #661, opened 2018-07-18
Ensure `runLocally` test resets environment variable.

bmaher:
This PR fixes a bug where tests were running with `ZALENIUM_RUNNING_LOCALLY` set to `true` and causing failures in some environments (CDP).

### Description
The reason for this environment variable being set to `true` is that the test `runLocally` in `ZaleniumRegistryTest.java` explicitly sets it but doesn't clear its value after the test.

This PR moves the `runLocally` test to its own file with explicit setting of the environment variable in `setUp` and `tearDown` methods. It also ensures that the environment variable is loaded into Zalenium's configuration correctly after changing it.

### Motivation and Context
Fix failure in CDP.

### How Has This Been Tested?
Ran `mvn clean test` with the code changes in a Vagrant copy of CDP's Build VM.

### Types of changes

### Checklist:

### DIFF
--- a/src/test/java/de/zalando/ep/zalenium/registry/ZaleniumRegistryTest.java
+++ b/src/test/java/de/zalando/ep/zalenium/registry/ZaleniumRegistryTest.java
@@ -73,7 +73,7 @@ public class ZaleniumRegistryTest {
 
         try {
             registry.add(p1);
-            
+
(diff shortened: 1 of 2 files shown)


bmaher:
__N.B.__ Investigation as to exactly why the environment variables behave differently on different environments is still underway.

diemol:
That's interesting... thanks for looking into that @bmaher!

diemol:
👍

bmaher:
👍

PR #690, opened 2018-08-09
cpuLimit and memoryLimit added for selenium desired containers

furkandogan:
Hi @diemol;
I open PR new again.
I develop limit resources parameter for selenium containers.

We can run that script "sudo docker run --rm -ti --name zalenium -d -p 4444:4444 -v /var/run/docker.sock:/var/run/docker.sock -v /tmp/zalenium:/home/seluser/videos --privileged zalenium:3.13.0c-SNAPSHOT start --desiredContainers 2 --maxDockerSeleniumContainers 2 --maxTestSessions 2 --screenWidth 1920 --screenHeight 1080 --timeZone "Europe/Istanbul" --videoRecordingEnabled true --keepOnlyFailedTests false --seleniumImageName 192.168.10.88:8123/elgalu-selenium --cpuLimit 20000000 --memoryLimit 536870912"

cpu value  20000000 = 0.2 cpu 
memory value 536870912 = 512mb

default cpu value 100000000 = 1 cpu
default memory value 1073741824 = 1GB

### DIFF
--- a/scripts/zalenium.sh
+++ b/scripts/zalenium.sh
@@ -7,6 +7,8 @@ CHROME_CONTAINERS=1
 FIREFOX_CONTAINERS=1
 DESIRED_CONTAINERS=${DESIRED_CONTAINERS:-2}
 MAX_DOCKER_SELENIUM_CONTAINERS=${MAX_DOCKER_SELENIUM_CONTAINERS:-10}
+SELENIUM_CONTAINER_CPU_LIMIT=${SELENIUM_CONTAINER_CPU_LIMIT:100000000}
+SELENIUM_CONTAINER_MEMORY_LIMIT=${SELENIUM_CONTAINER_MEMORY_LIMIT:1073741824}
 ZALENIUM_ARTIFACT="$(pwd)/${project.build.finalName}.jar"
 DEPRECATED_PARAMETERS=false
 SAUCE_LABS_ENABLED=${SAUCE_LABS_ENABLED:-false}
@@ -354,6 +356,8 @@ StartUp()
     fi
     export ZALENIUM_DESIRED_CONTAINERS=${DESIRED_CONTAINERS}
     export ZALENIUM_MAX_DOCKER_SELENIUM_CONTAINERS=${MAX_DOCKER_SELENIUM_CONTAINERS}
+    export ZALENIUM_SELENIUM_CONTAINER_CPU_LIMIT=${SELENIUM_CONTAINER_CPU_LIMIT}
+    export ZALENIUM_SELENIUM_CONTAINER_MEMORY_LIMIT=${SELENIUM_CONTAINER_MEMORY_LIMIT}
     export ZALENIUM_VIDEO_RECORDING_ENABLED=${VIDEO_RECORDING_ENABLED}
     export ZALENIUM_TZ=${TZ}
     export ZALENIUM_SCREEN_WIDTH=${SCREEN_WIDTH}
@@ -571,7 +575,7 @@ StartUp()
             --maxTestSessions $MAX_TEST_SESSIONS --sauceLabsEnabled $SAUCE_LABS_ENABLED
             --browserStackEnabled $BROWSER_STACK_ENABLED --testingBotEnabled $TESTINGBOT_ENABLED
             --videoRecordingEnabled $VIDEO_RECORDING_ENABLED --screenWidth $SCREEN_WIDTH --screenHeight $SCREEN_HEIGHT
-            --timeZone $TZ"
+            --timeZone $TZ --cpuLimit $SELENIUM_CONTAINER_CPU_LIMIT --memoryLimit $SELENIUM_CONTAINER_MEMORY_LIMIT"
 
         local args=(
             --max-time 10
@@ -758,6 +762,8 @@ function usage()
     echo -e "\t --gridPassword -> allows you to specify a password to enable basic auth protection, --gridUser must be provided also."
     echo -e "\t --maxTestSessions -> max amount of tests executed per container, defaults to '1'."
     echo -e "\t --keepOnlyFailedTests -> Keeps only videos of failed tests (you need to send a cookie). Defaults to 'false'"
+    echo -e "\t --cpuLimit -> Selenium Container Cpu Limit (you need to send a cookie). Defaults to '100000000' 1 CPU"
+    echo -e "\t --memoryLimit -> Selenium Container Memory Limit (you need to send a cookie). Defaults to '1073741824' is 1GB"
     echo ""
     echo -e "\t stop"
     echo ""
@@ -801,6 +807,12 @@ case ${SCRIPT_ACTION} in
                     ;;
                 --maxDockerSeleniumContainers)
                     MAX_DOCKER_SELENIUM_CONTAINERS=${VALUE}
+                    ;;
+                --cpuLimit)
+                    SELENIUM_CONTAINER_CPU_LIMIT=${VALUE}
+                    ;;
+                --memoryLimit)
+                    SELENIUM_CONTAINER_MEMORY_LIMIT=${VALUE}
                     ;;
                 --sauceLabsEnabled)
                     SAUCE_LABS_ENABLED=${VALUE}
--- a/src/main/java/de/zalando/ep/zalenium/container/ContainerClient.java
+++ b/src/main/java/de/zalando/ep/zalenium/container/ContainerClient.java
@@ -5,7 +5,7 @@ import java.net.URL;
 import java.util.Map;
 
 public interface ContainerClient {
-    
+
     void setNodeId(String nodeId);
 
     ContainerClientRegistration registerNode(String zaleniumContainerName, URL remoteHost);
@@ -22,11 +22,13 @@ public interface ContainerClient {
 
     ContainerCreationStatus createContainer(String zaleniumContainerName, String image, Map<String, String> envVars, String nodePort);
 
+    ContainerCreationStatus createContainer(String zaleniumContainerName, String image, String cpuLimit, String memoryLimit, Map<String, String> envVars, String nodePort);
+
     void initialiseContainerEnvironment();
 
     String getContainerIp(String containerName);
-    
+
     boolean isReady(ContainerCreationStatus container);
-    
+
     boolean isTerminated(ContainerCreationStatus container);
 }
--- a/src/main/java/de/zalando/ep/zalenium/container/DockerContainerClient.java
+++ b/src/main/java/de/zalando/ep/zalenium/container/DockerContainerClient.java
@@ -1,25 +1,5 @@
 package de.zalando.ep.zalenium.container;
 
-import java.io.IOException;
-import java.io.InputStream;
-import java.net.URL;
-import java.nio.charset.StandardCharsets;
-import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.HashMap;
-import java.util.List;
-import java.util.Map;
-import java.util.Objects;
-import java.util.Optional;
-import java.util.concurrent.atomic.AtomicBoolean;
-import java.util.stream.Collectors;
-
-import com.spotify.docker.client.messages.PortBinding;
-import org.apache.commons.io.IOUtils;
-import org.apache.commons.lang3.RandomStringUtils;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
-
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.collect.ImmutableMap;
 import com.spotify.docker.client.AnsiProgressHandler;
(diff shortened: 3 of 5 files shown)


scripts/zalenium.sh
```diff
@@ -7,6 +7,8 @@ CHROME_CONTAINERS=1
 FIREFOX_CONTAINERS=1
 DESIRED_CONTAINERS=${DESIRED_CONTAINERS:-2}
 MAX_DOCKER_SELENIUM_CONTAINERS=${MAX_DOCKER_SELENIUM_CONTAINERS:-10}
+SELENIUM_CONTAINER_CPU_LIMIT=${SELENIUM_CONTAINER_CPU_LIMIT:100000000}
```
pearj: I don't think there should be a default, as it should be disabled if not specified.

src/main/java/de/zalando/ep/zalenium/container/kubernetes/KubernetesContainerClient.java
```diff
@@ -1,16 +1,24 @@
 package de.zalando.ep.zalenium.container.kubernetes;
 
+import de.zalando.ep.zalenium.container.ContainerClient;
```
pearj: Please revert all changes to this file. Also kubernetes already has its own incompatible notion of cpu limits and requests.

src/main/java/de/zalando/ep/zalenium/container/ContainerClient.java
```diff
@@ -22,11 +22,13 @@
 
     ContainerCreationStatus createContainer(String zaleniumContainerName, String image, Map<String, String> envVars, String nodePort);
 
+    ContainerCreationStatus createContainer(String zaleniumContainerName, String image, String cpuLimit, String memoryLimit, Map<String, String> envVars, String nodePort);
```
pearj: I don't think we should add the cpu limits this way. I think it might make more sense to import the system variable directly into the docker implementation, or extend the kubernetes implementation to support docker.

https://github.com/zalando/zalenium/blob/master/src/main/java/de/zalando/ep/zalenium/container/kubernetes/KubernetesContainerClient.java#L495

Specifying `100000000` as one cpu seems crazy, maybe we should use the kubernetes definition? And find a way to convert the value:

https://kubernetes.io/docs/concepts/configuration/manage-compute-resources-container/#meaning-of-cpu

scripts/zalenium.sh
```diff
@@ -7,6 +7,8 @@ CHROME_CONTAINERS=1
...
+SELENIUM_CONTAINER_MEMORY_LIMIT=${SELENIUM_CONTAINER_MEMORY_LIMIT:1073741824}
```
pearj: Same goes here, there is no sensible default, as every web app is different.

scripts/zalenium.sh
```diff
@@ -354,6 +356,8 @@ StartUp()
     fi
     export ZALENIUM_DESIRED_CONTAINERS=${DESIRED_CONTAINERS}
     export ZALENIUM_MAX_DOCKER_SELENIUM_CONTAINERS=${MAX_DOCKER_SELENIUM_CONTAINERS}
+    export ZALENIUM_SELENIUM_CONTAINER_CPU_LIMIT=${SELENIUM_CONTAINER_CPU_LIMIT}
```
diemol: I don't think it is necessary to change this file, the idea would be just to have the cpuLimit and memoryLimit as environment variables and use them when they are available.

Please revert the changes to this file and just grab the env vars in the Java code.

src/main/java/de/zalando/ep/zalenium/container/ContainerClient.java
diemol: I agree with @pearj, no need change this methods, a more simple approach is to get the env vars in the `DockerContainerClient.java` class.

src/main/java/de/zalando/ep/zalenium/container/DockerContainerClient.java
```diff
@@ -29,23 +9,27 @@
...
-import com.spotify.docker.client.messages.ContainerConfig;
-import com.spotify.docker.client.messages.ContainerCreation;
-import com.spotify.docker.client.messages.ContainerInfo;
-import com.spotify.docker.client.messages.ContainerMount;
-import com.spotify.docker.client.messages.ExecCreation;
-import com.spotify.docker.client.messages.HostConfig;
-import com.spotify.docker.client.messages.Image;
-import com.spotify.docker.client.messages.NetworkSettings;
-
+import com.spotify.docker.client.messages.*;
```
diemol: Don't optimise imports please, leave every import explicit.

src/main/java/de/zalando/ep/zalenium/container/DockerContainerClient.java
```diff
@@ -29,23 +9,27 @@
...
+import org.apache.commons.io.IOUtils;
+import org.apache.commons.lang3.RandomStringUtils;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.io.IOException;
+import java.io.InputStream;
+import java.net.URL;
+import java.nio.charset.StandardCharsets;
+import java.util.*;
```
diemol: Same here, please don't optimise imports.

src/main/java/de/zalando/ep/zalenium/container/DockerContainerClient.java
```diff
@@ -229,12 +213,21 @@ public int getRunningContainers(String image) {
         return 0;
     }
 
-    public ContainerCreationStatus createContainer(String zaleniumContainerName, String image, Map<String, String> envVars,
+    @Override
+    public ContainerCreationStatus createContainer(String zaleniumContainerName, String image, Map<String, String> envVars, String nodePort) {
+        return createContainer(zaleniumContainerName, image, DEFAULT_SELENIUM_CONTAINER_CPU_LIMIT, DEFAULT_SELENIUM_CONTAINER_MEMORY_LIMIT, envVars, nodePort, NAME_COLLISION_RETRIES);
+    }
+
+    public ContainerCreationStatus createContainer(String zaleniumContainerName, String image,
```
diemol: There is no need to change the method signature, just grab the environment variables inside it, either by calling a method or just with a couple of code lines.

src/main/java/de/zalando/ep/zalenium/container/DockerContainerClient.java
```diff
@@ -272,12 +265,14 @@ public ContainerCreationStatus createContainer(String zaleniumContainerName, Str
...
-            .extraHosts(extraHosts)
-            .autoRemove(true)
-            .storageOpt(storageOpt)
-            .privileged(isZaleniumPrivileged);
+                .appendBinds(binds)
+                .networkMode(networkMode)
+                .extraHosts(extraHosts)
+                .autoRemove(true)
+                .storageOpt(storageOpt)
+                .nanoCpus(Long.valueOf(cpuLimit))
```
diemol: Please only add this if the environment variables are present, we should not have defaults since every user is different. Just nest it into an "if".

src/main/java/de/zalando/ep/zalenium/container/kubernetes/KubernetesContainerClient.java
diemol: I agree, this file does not need to be touched by this PR.

src/main/java/de/zalando/ep/zalenium/proxy/DockeredSeleniumStarter.java
```diff
@@ -22,13 +17,9 @@
...
 import org.openqa.selenium.remote.CapabilityType;
 
-import com.google.common.annotations.VisibleForTesting;
-
-import de.zalando.ep.zalenium.container.ContainerClient;
-import de.zalando.ep.zalenium.container.ContainerCreationStatus;
-import de.zalando.ep.zalenium.container.ContainerFactory;
-import de.zalando.ep.zalenium.matcher.ZaleniumCapabilityType;
-import de.zalando.ep.zalenium.util.Environment;
+import java.util.*;
```
diemol: Please revert this part, all imports should be explicit.

src/main/java/de/zalando/ep/zalenium/proxy/DockeredSeleniumStarter.java
```diff
@@ -56,6 +47,10 @@
     private static final Logger LOGGER = Logger.getLogger(DockeredSeleniumStarter.class.getName());
     private static final String DEFAULT_DOCKER_SELENIUM_IMAGE = "elgalu/selenium";
     private static final String ZALENIUM_SELENIUM_IMAGE_NAME = "ZALENIUM_SELENIUM_IMAGE_NAME";
+    public static final String DEFAULT_SELENIUM_CONTAINER_CPU_LIMIT = "100000000";
```
diemol: Let's have the env vars name declared but without defaults. This part can all be done in the `DockerContainerClient.java` class. Basically, I think there is no need to touch this class either.  Can you please revert the changes here?

scripts/zalenium.sh
furkandogan: In kubernetes module there are cpuLimit and memLimit values. So I think it should have default value. And when we use without default it would be load server's property so that server could be shutdown. So ı developed like this @pearj

scripts/zalenium.sh
furkandogan: if ı don't use in this sh files. When I run that script "sudo docker run --rm -ti --name zalenium -d -p 4444:4444 -v /var/run/docker.sock:/var/run/docker.sock -v /tmp/zalenium:/home/seluser/videos --privileged zalenium:3.13.0c-SNAPSHOT start --cpuLimit 20000000 --memoryLimit 536870912",  it wouldn't work.

scripts/zalenium.sh
diemol: It is true that the Kubernetes part has values, but it was developed like that since day 1. The Docker part does not have any default values, and introducing them is not backwards compatible since we don't know what a good default is for all people who are using Zalenium with docker. 

If we arbitrarily set a default value, it is very probable that it will affect a bunch of people and many issues about that will be created, which will increase the maintenance time for the project.

To make it short: yes, technically it makes sense to have defaults, but we are not going to introduce them because this would affect more users than actually benefit them.

scripts/zalenium.sh
diemol: It is possible to pass them as environment variables, e.g.:

```
zalenium -d -p 4444:4444 -v /var/run/docker.sock:/var/run/docker.sock 
-v /tmp/zalenium:/home/seluser/videos 
-e ZALENIUM_SELENIUM_CONTAINER_CPU_LIMIT=20000000
-e ZALENIUM_SELENIUM_CONTAINER_MEMORY_LIMIT=536870912
--privileged zalenium:3.13.0c-SNAPSHOT start ",
```
So there is no need to touch the zalenium.sh script.

scripts/zalenium.sh
furkandogan: Ok, @diemol thanks for your review. I will change this request

furkandogan:
@pearj and @diemol I changed code for your requests.

diemol:
Looks much better, thanks! I will add some new comments.

src/main/java/de/zalando/ep/zalenium/container/ContainerClient.java
```diff
@@ -5,7 +5,7 @@
 import java.util.Map;
 
 public interface ContainerClient {
-
```
diemol: Please revert changes to this file, it is just some spaces and tabs that changed.

src/main/java/de/zalando/ep/zalenium/container/kubernetes/KubernetesContainerClient.java
```diff
@@ -109,7 +109,7 @@ public KubernetesContainerClient(Environment environment,
                             + "\tSelenium Pod Resource Limits: %s\n"
                             + "\tSelenium Pod Resource Requests: %s",
                     hostname, appName, zaleniumAppName,
-                            seleniumPodLimits.toString(), seleniumPodRequests.toString()));
```
diemol: Please revert changes to this file, it is just some spaces and tabs that changed.

src/main/java/de/zalando/ep/zalenium/proxy/DockeredSeleniumStarter.java
```diff
@@ -79,7 +79,7 @@
             "zalenium_https_proxy",
             "zalenium_no_proxy"
     };
-
```
diemol: Please revert changes to this file, it is just some spaces and tabs that changed.

src/main/java/de/zalando/ep/zalenium/container/DockerContainerClient.java
```diff
@@ -272,12 +306,20 @@ public ContainerCreationStatus createContainer(String zaleniumContainerName, Str
...
+                .networkMode(networkMode)
+                .extraHosts(extraHosts)
+                .autoRemove(true)
+                .storageOpt(storageOpt)
+                .privileged(isZaleniumPrivileged);
+
+        String cpuLimit = getSeleniumContainerCpuLimit();
+        String memoryLimit = getSeleniumContainerMemoryLimit();
+
+        if (cpuLimit != null)
```
diemol: If the env var is not set, does it really come back as null? Or does it come back as an empty string?

Perhaps you can use `Strings.isNullOrEmpty` from google.commons to cover both cases.

src/main/java/de/zalando/ep/zalenium/container/DockerContainerClient.java
```diff
@@ -272,12 +306,20 @@ public ContainerCreationStatus createContainer(String zaleniumContainerName, Str
...
+            hostConfigBuilder.nanoCpus(Long.valueOf(cpuLimit));
+        if(memoryLimit != null)
```
diemol: Same here.

pearj:
I think the commits would need to be squashed if this were going to be merged, so as to not pollute the commit history

diemol:
Ah ok, I did not notice the last commit that solved the last comments. We will test it, thanks @furkandogan!

PR #708, opened 2018-08-28
fix(cleanup): avoid marking nodes stale while they are cleaning up

### LINKED ISSUE #579: Duplicate entries in dashboard
Zalenium Image Version(s):  
3.11.0g
Docker Version:  
Openshift origin 3.7

Zalenium is started with json logging enabled 

## Expected Behavior -
One entry per test result in dashboard

## Actual Behavior -
For tests that are flagged as "timeout", sometime the same result shows up twice in the dashboard

Example of logs showing copying the video multiple times:
image

Also, it seems like every copying video log message is happening twice per copy (might be introduced by the json logging that was added).

Also, for these test results that show up as duplicates, the Log -> Selenium log and Webdriver Log is always missing even though I see this event in the logs: `zalenium-40000-b7kds completed [bash, -c, transfer-logs.sh]`

ecktom:
### Description
Fixes #579 

### How Has This Been Tested?
Normal behaviour was tested locally and still works.

### Types of changes

### Checklist:


diemol:
Looks fine for me too, thanks @ecktom!
I will merge this into a protected branch so we can run all the Travis tests.

PR #728, opened 2018-09-27
Fix prometheus bug where running/starting containers could be wrong

pearj:
### Description
In some circumstances, the Prometheus metrics `selenium_containers_starting` and `selenium_containers_running` could be inaccurate.  This fix changes the way that the data is collected.
### Motivation and Context
Previously the metric was a normal gauge incremented and decremented, but it seems there were some edge cases that weren't handled.  This fix directly reads the map, which is the source of truth, so if the metric is inaccurate in the future, then that would be an indication of a bug inside the `de.zalando.ep.zalenium.proxy.AutoStartProxySet.AutoStartProxySet` class.

As part of this change these metrics: `selenium_containers_starting` and `selenium_containers_running`.
Have been replaced with:
`selenium_containers{state="running"}` and `selenium_containers{state="starting"}` as this lines up with how metrics are usually organised in Prometheus.

### How Has This Been Tested?
Tested it manually in my local Minishift and started some containers and watched the results of `/metrics`.

Additionally, created a test to cover the Prometheus Collector class.

### Types of changes
- [x] Breaking change (fix or feature that would cause existing functionality to change)

### Checklist:

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -342,6 +342,7 @@
             <plugin>
                 <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-jar-plugin</artifactId>
+                <version>2.4</version>
                 <configuration>
                     <archive>
                         <manifestSections>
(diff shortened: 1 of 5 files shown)


pom.xml
```diff
@@ -342,6 +342,7 @@
             <plugin>
                 <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-jar-plugin</artifactId>
+                <version>2.4</version>
```
pearj: This is to get rid of the warning that maven complains about when building zalenium

diemol: 👍

diemol:
Thanks @pearj!

PR #744, opened 2018-10-12
Adding basic authentication values in K8s helm chart

### LINKED ISSUE #743: Ability to use basic auth in helm deployed K8s zalenium
Even though it's probably not the *most* secure way of authentication, the basic auth available from Zalenium, provides added security to any deployment.  The helm chart found in this repository does not provide a way to pass in the required basic authentication parameters to the zalenium container, nor does it provide a way to adjust the readiness/liveness probe to take into account the authentication.

I've already done the work for this, but am creating an issue as per the contributing guidelines.

If using Kubernetes, specify your environment, and if relevant your manifests:  helm deployed

## Expected Behavior -
There should be values available in the helm chart values that will set up basic authentication on a helm deployed k8s instance of Zalenium while maintaining all existing functionality including installs without basic auth.

## Actual Behavior -
Does not currently exist.

wlaoh:
Resolves #743 

### Description

I've added additional functionality into the helm chart which allows the user to specify if they want to use basic authentication and the credentials they want to use.  I've also expanded the information that is provided to the user after a successful helm install because "Hi" just didn't seem as cool.

I've updated the readiness/liveness probe to take into account the usage of basic authentication.  Created a template helper that would create a base64 encoded value of the username and password to be used by the readiness/liveness probe http headers.

### Motivation and Context

I wanted this, because I love being able to deploy Zalenium using helm into my K8s cluster, but wasn't happy that there wasn't an easy way to get basic authentication working.  In today's security climate I think the least we can do is put up a username and password in front of it.

### How Has This Been Tested?

I tested this by helm linting the entire chart then I performed the following test cases:
1. Installed this chart using different user credentials from the defaults
2. Installed this chart using the default user credentials
3. Installed this chart without any basic auth

All test cases passed.  I was able to reach all pages (dashboard, console, etc.)

### Types of changes

### Checklist:

### DIFF
--- a/docs/k8s/helm/README.md
+++ b/docs/k8s/helm/README.md
@@ -81,8 +81,11 @@ See Zalenium's [usage examples](https://github.com/zalando/zalenium/blob/master/
 | `hub.sauceAccessKey` | Access key to log into saucelabs | blank |
 | `hub.browserStackUser` | Credentials for browserstack | blank |
 | `hub.browserStackKey` | Credentials for browserstack | blank |
-| `hub.testingBotKey` | Credentials for testingbot | blank | 
-| `hub.testingBotSecret` | Credentials for testingbot | blank | 
+| `hub.testingBotKey` | Credentials for testingbot | blank |
+| `hub.testingBotSecret` | Credentials for testingbot | blank |
+| `hub.basicAuth.enabled` | Enables basic authentication | false |
+| `hub.basicAuth.username` | Username for basic authentication | zalenium |
+| `hub.basicAuth.password` | Credentials for testingbot | password |
 
 Specify each parameter using the `--set key=value[,key=value]` argument to `helm install`. For example,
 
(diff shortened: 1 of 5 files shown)


docs/k8s/helm/README.md
```diff
@@ -81,8 +81,11 @@ See Zalenium's [usage examples](https://github.com/zalando/zalenium/blob/master/
 | `hub.sauceAccessKey` | Access key to log into saucelabs | blank |
 | `hub.browserStackUser` | Credentials for browserstack | blank |
 | `hub.browserStackKey` | Credentials for browserstack | blank |
-| `hub.testingBotKey` | Credentials for testingbot | blank | 
-| `hub.testingBotSecret` | Credentials for testingbot | blank | 
+| `hub.testingBotKey` | Credentials for testingbot | blank |
+| `hub.testingBotSecret` | Credentials for testingbot | blank |
+| `hub.basicAuth.enabled` | Enables basic authentication | false |
+| `hub.basicAuth.username` | Username for basic authentication | zalenium |
+| `hub.basicAuth.password` | Credentials for testingbot | password |
```
diemol: There is a typo here, it should not be "credentials for testingbot", but something else.

wlaoh: @diemol fixed, thanks for catching.

diemol:
Thanks!

PR #764, opened 2018-11-07
Cache the selenium-node image in cluster

cbll:
### Description

Sets the `imagePullPolicy` to `IfNotPresent` instead of `Always`. 
This should allow caching in the cluster so that k8s does not have to make a request to the docker hub every time a node is to be created, thereby lowering overhead and network IO.
Apart from this, I have observed that scaling up nodes might present issues for docker hub - likely resource constraints from their side.

To avoid readiness probe errors from the nodes, it could be beneficial setting up the timeout from 1 second to 5. There isn't a lot of room for error with just one second; at the least, it's been observed that this made it slightly more stable from our side.

### Motivation and Context

While experimenting with zalenium moving from a vanilla selenium grid, we have observed some issues scaling up the amount of workers(nodes). Both in terms of readiness probe for the nodes, as well as receiving errors from connecting the individual nodes to docker hub to fetch the `selenium` image.

By changing the `imagePullPolicy` to `IfNotPresent`, we should be able to cache one version and pull it locally. If it's hard-set to `Always`, the amount of requests and downloads of images from docker-hub increases in relation to the amount of nodes, which isn't ideal for scalability.

### How Has This Been Tested?

I pulled down the project, changed the values, pushed the `zalenium` image to our local image repository and made it a dependency of a `zalenium` deployment we have. After doing so, I observed an increase in speed of executing tests(due to less networking overhead) as well as stability(since Docker hub is no longer a hard dependency of node creation).

### Types of changes

- [X ] Bug fix (non-breaking change which fixes an issue)

### Checklist:

- [X ] My code follows the code style of this project.
- [ X] I have read the **CONTRIBUTING** document.
- [ X] All new and existing tests passed.

I am not a Kubernetes expert; but, from talking to some that are much more knowledgeable than myself on the subject - it would sound as if setting `imagePullPolicy` to `Always` is generally not considered best practice. However, happy to learn why that might not be the case here! :-)


src/main/java/de/zalando/ep/zalenium/container/kubernetes/KubernetesContainerClient.java
```diff
@@ -577,7 +591,7 @@ public static DoneablePod createDoneablePodDefaultImpl(PodConfiguration config)
                             .withInitialDelaySeconds(5)
                             .withFailureThreshold(60)
                             .withPeriodSeconds(1)
-                            .withTimeoutSeconds(1)
+                            .withTimeoutSeconds(5)
```
pearj: Is your environment slow enough, that you're never passing a health check?

pearj:
With the `Always` setting it's not actually downloading the entire image every time a node starts.  It's just checking in with Docker Hub to see if the image has changed, if not then only tag information is downloaded.

By using `IfNotPresent` that means you'll never get updates, and you could get inconsistent versions of selenium images across your Kubernetes nodes if you're leaving it to the default of `elgalu/selenium:latest`, if you use a specific version like `elgalu/selenium:3.12.0-p10`, then that'd be safe.

In my case, we use a private registry built into OpenShift instead of using docker hub, then you get the best of both worlds, it downloads quickly and you can get consistent versions by using latest still.

src/main/java/de/zalando/ep/zalenium/container/kubernetes/KubernetesContainerClient.java
pearj: Are you using video recording?

pearj: How large are your k8s nodes in cpu and memory?

cbll:
> With the `Always` setting it's not actually downloading the entire image every time a node starts. It's just checking in with Docker Hub to see if the image has changed, if not then only tag information is downloaded.

Fair enough. Regardless, I started seeing a bunch of errors - both `503` thrown back from docker hub as well as access denied; I guess from too many requests with a short timespan. Obviously, this is solved by using a private registry, but still doesn't seem ideal for something that scales up to a relatively large number pretty fast. Though, this issue may be on docker hub, I suppose. 

> By using `IfNotPresent` that means you'll never get updates, and you could get inconsistent versions of selenium images across your Kubernetes nodes if you're leaving it to the default of `elgalu/selenium:latest`, if you use a specific version like `elgalu/selenium:3.12.0-p10`, then that'd be safe.

I have, indeed, hard-coded the selenium image to `elgalu/selenium:3.14.0-p14` and thereby managing it by hand to  maintain consistency.

> In my case, we use a private registry built into OpenShift instead of using docker hub, then you get the best of both worlds, it downloads quickly and you can get consistent versions by using latest still.

The end-game is something similar to this for our setup. But, we currently can't get a test suite to run on zalenium without it falling over with a `503 Unavailable` at the moment. It does, however, recover fast - but it obviously puts a pretty big element of uncertainty on a full run.

src/main/java/de/zalando/ep/zalenium/container/kubernetes/KubernetesContainerClient.java
cbll: Video recording is currently switched off.

Here are the resources for a given node fetched via `kubectl get pod zalenium-40000-<UID>`

```   resources:
      limits:
        cpu: "1"
      requests:
        cpu: 250m
        memory: 4Gi
```

src/main/java/de/zalando/ep/zalenium/container/kubernetes/KubernetesContainerClient.java
```diff
@@ -557,7 +571,7 @@ public static DoneablePod createDoneablePodDefaultImpl(PodConfiguration config)
                     .addNewContainer()
                         .withName("selenium-node")
                         .withImage(config.getImage())
-                        .withImagePullPolicy("Always")
+                        .withImagePullPolicy(ImagePullPolicy.IfNotPresent.getPolicy())
```
pearj: It should default to `Always`, but allow overriding to `IfNotPresent` otherwise you'll silently break other people relying on rolling tags.

src/main/java/de/zalando/ep/zalenium/container/kubernetes/KubernetesContainerClient.java
pearj: Fair enough. 5 seconds is ok, I can live with that default.

src/main/java/de/zalando/ep/zalenium/container/kubernetes/KubernetesContainerClient.java
cbll: Gotcha, so make a field on `PodConfiguration.java` that includes an `imagePullPolicy` field that then defaults to `Always` in `KubernetesContainerClient.java`, unless overridden..

Not sure if I understand the whole flow after that; how would you map a given parameter that I(as a "user" of zalenium) provides it? Where would I practically override such variable from?

cbll: I added another commit - hope I'm in the right direction(and that Windows+IntelliJ didn't mess up formatting too much).. I am unsure of how to actually read the environment variable(I guess, an enum similarly to the Resources one?) - Any help here would be appreciated. It passes the tests as-is, but obviously defaults to imagePullPolicy "Always" - always 😄. My Java is rusty, to be frank!

cbll: - Committed a suggestion of how to do it. As said, my Java is a bit rusty, so feel free to rip it apart 👍

src/main/java/de/zalando/ep/zalenium/container/kubernetes/KubernetesContainerClient.java
```diff
@@ -139,6 +140,14 @@ private void buildResourceMaps() {
                 }
             }
         }
+
+        String imagePullPolicyEnvValue = environment.getStringEnvVariable("ZALENIUM_KUBERNETES_IMAGE_PULL_POLICY", null);
```
diemol: Maybe you could use `Optional.ofNullable` here, just to make it a bit more clean. [Here](https://github.com/zalando/zalenium/blob/master/src/main/java/de/zalando/ep/zalenium/proxy/DockeredSeleniumStarter.java#L160) is an example.

diemol:
Hi @cbll, thanks for this PR! I just made a small comment.

src/main/java/de/zalando/ep/zalenium/container/kubernetes/KubernetesContainerClient.java
cbll: Sorry about the whitespaces, something went funky when I switched from ubuntu to windows and copied the project -_- I'll fix it in the last commit.

Can you elaborate on the above - shouldn't it be `Optional.Of` for a String? Or perhaps, I am misunderstanding. Can you give an example of what you'd change it to?

diemol: No worries! Thanks for fixing the formatting.

I think I just made it more complicated with my comment, reading it a second time, perhaps this is more simple:

```java
String imagePullPolicyEnvValue = environment.getStringEnvVariable("ZALENIUM_KUBERNETES_IMAGE_PULL_POLICY", ImagePullPolicyType.Always.name());
```

Since the `getStringEnvVariable` will return the second parameter when the first one is `null`. Then, the if-else is not needed. What do you think?

diemol:
Looks good, thanks @cbll!

diemol:
@cbll if you have time, would be cool if you can send a PR documenting the change and also updating the Helm chart :)

cbll:
@diemol sure thing - where's the helm chart for pods at? I can only find the one for the hub, it appears(I don't know much about Helm, tbh :-))

https://github.com/zalando/zalenium/tree/master/docs/k8s/helm

pearj:
@cbll Pods are created on demand by the hub, which is the whole point of zalenium.

cbll:
@pearj I am aware - I'd love to update the helm chart, but I don't know what would go in the above which I linked to, since it seems to set configuration of `hub.something` exclusively :-)

pearj:
@cbll Oh right I see what you mean.  What @diemol is asking is that you add the environment variable to the helm chart for the hub, so that other people using helm can easily use your feature.

diemol:
@cbll yes, that is what I meant, sorry if I was not specific enough.

Issue #631, opened 2018-06-27
Dashboard is not showing - k8s


dusansusic:
Zalenium Image Version(s):  
3.12e
Docker Version:  
17.3.2
OS: 
ubuntu 16.04

I found an issue here: https://github.com/zalando/zalenium/issues/353 and it aims that's solved but it's not.

When i deployed zalenium to the k8s, dashboard si not available till I log in into container:
`kubectl exec -it zalenium-799b7bff48-45ktv bash -n infra`
and execute:
`sudo cp * videos/`
it will work.

I am using k8s 1.10.3 accessing to dashboard:
`http://IP:PORT/dashboard`
or
`http://IP:PORT/dashboard/` - problem exists on both ways

```
          volumeMounts:
            - name: zalenium-data
              mountPath: /home/seluser/videos
```

Should we use `/home/seluser`, instead of `/home/seluser/videos`?

Also, Service is not accessible at: `http://IP:4444` but `http://IP:PORT`, eg. `http://IP:32152`

Service definition is identical to your:
```
apiVersion: v1
kind: Service
metadata:
  labels:
    app: zalenium
  name: zalenium
  namespace: infra
spec:
  ports:
  - name: zalenium
    port: 4444
    targetPort: 4444
  selector:
    app: zalenium
    role: grid
  type: NodePort
```

dusansusic:
Anyone else who experience this issue with dashboard?

pearj:
Port `4444` is the port that zalenium uses inside the container. Your nodeport is completely different, generally nodeports aren't useful for the internet anyhow. As most people would be using some sort of ingress.

On startup, it automatically tries to copy the dashboard files to the /home/seluser/videos mount.  Maybe look for copy errors on zalenium pod start, maybe you have some sort of permissions issues on your volume mount.

dusansusic:
@pearj  Thanks for reply.

Well, ignore the parts with ports, my bad.

Volume is provisioned as EBS on AWS, I am not mounting it at all, k8s do that.
Container path is `/home/seluser/videos` and mount point on node is `/tmp/videos`.
I have permission issues:
```Kubernetes service account found.
Copying files for Dashboard...
cp: cannot create regular file '/home/seluser/videos/index.html': Permission denied
cp: cannot create directory '/home/seluser/videos/css': Permission denied
cp: cannot create directory '/home/seluser/videos/js': Permission denied
```
What I need to do to fix this? ^


also, I have RBAC enabled cluster and I created roles:
[...]

PR #772, opened 2018-11-18
standardize all k8s vanilla deployment

### LINKED ISSUE #750: Kubernetes and Openshift standardization
Hi,

Currently there are 2-3 ways to deploy `zalenium` on Openshift that I'm aware of: 

1. Apply the kubernetes dir https://github.com/zalando/zalenium/tree/master/kubernetes as documented [here](https://opensource.zalando.com/zalenium/#kubernetes)
1. Process and then apply the template https://github.com/zalando/zalenium/blob/master/docs/k8s/zalenium-template.yaml
1. Install with helm https://github.com/zalando/zalenium/tree/master/docs/k8s/helm

It would be nice if we can manage a helm template which is compatible with all Kubernetes vanilla and get rid of the others.

To be compatible with all Kubernetes vanilla, it means considering the default policy in Openshift: 

1. A pod cannot run as root. It is started with a random user id part of the root user group (0).
1. A pod cannot run with privilege.

From what I know, it should be possible to follow these 2 guidelines in `zalenium`.

Also for users that do not want `tiller` which is coming with `helm v2`, we can document the alternative of 

```
helm template --set foo=bar docs/k8s/helm/local/zalenium | kubectl apply -n <target_namespace> -f -
``` 

which is the equivalent of 

```
oc process -p foo=bar -f docs/k8s/zalenium-template.yaml | oc apply -n <target_namespace> -f -
```

I tried the current helm template in openshift and it is not working because of the init container https://github.com/zalando/zalenium/blob/master/docs/k8s/helm/local/zalenium/templates/deployment.yaml#L24-L29 (which is not defined in the openshift template).

I'm not sure why this is necessary especially because it comes from a PersitentVolume so it should have the appropriate rights already.

I think this would simplify life of users like that tries to find way to install `zalenium` in kubernetes.

What do you think ?

And then why not eventually publish this helm chart to the [official repo](https://github.com/helm/charts/) ?

arnaud-deprez:
### Description
The purpose of this PR is to fix #750 by using [Helm](https://helm.sh/) template for all kubernetes vanilla deployments.

[Helm](https://helm.sh/) has become the standard package manager running atop Kubernetes and so it seems reasonable to support only that type of deployment.

`Helm 2.x` typically uses a server side component called tiller which is responsible to create your Kubernetes objects and so it does not use the RBAC configuration set for the user but it uses the one set for tiller (more information on the [jenkins-x website](https://jenkins-x.io/news/helm-without-tiller))
Until `Helm 3` is GA, it is still possible to use `Helm 2` without `Tiller` by simply applying the result of the template `helm template --name zalenium charts/zalenium | kubectl apply -f -`

### Motivation and Context
There are currently many documented ways to deploy zalenium on vanilla kubernetes resulting in a maintenance overhead. 
Moreover some discrepancies between each deployment method have been identified in the issue.

### How Has This Been Tested?
- [x] Openshift (minishift) deployment with RBAC and PVs: `helm template --name zalenium --set ingress.enabled=true --set ingress.hostname=hub-zalenium.$(minishift ip).nip.io --set rbac.create=true --set serviceAccount.create=true --set persistence.enabled=true charts/zalenium | kubectl -n myproject apply -f -`
- [x] Kubernetes (minikube) deployment with RBAC and PVs: `helm template --name zalenium --set ingress.enabled=true --set ingress.hostname=hub-zalenium.$(minikube ip).nip.io --set rbac.create=true --set serviceAccount.create=true --set persistence.enabled=true charts/zalenium | kubectl -n zalenium apply -f -`
- [x] Kubernetes (minikube) deployment with RBAC, persistence and security context: `helm template --name zalenium --set ingress.enabled=true --set ingress.hostname=hub-zalenium.$(minikube ip).nip.io --set rbac.create=true --set serviceAccount.create=true --set persistence.enabled=true --set hub.securityContext.enabled=true charts/zalenium | kubectl -n zalenium apply -f -`

### Types of changes

### Checklist:


arnaud-deprez:
Hi @diemol @pearj,

The PR is finally ready to be merged.
FYI, I initially try to upgrade to latest `minikube` and `kubernetes` but the tests were failing so I left it for another eventual PR.

IMO, it is just working like that but if you have a way to try to reproduce #631 I can try but in both Minikube and Minishift, I can see the dashboard and the videos without setting any security context.
However this might be necessary in some cloud provider to use `fsGroup=0`.
But then I'm wondering if the Pods instanciated by the Hub shouldn't inherit this security context too.

diemol:
Hi @arnaud-deprez,

Thanks for taking the time to put this together. In general, this looks good to me as it simplifies the docs and maintenance. I just wonder if it makes sense to put the command examples you posted above to start in OpenShift and k8s. Also, if everyone is using helm now... I don't have a k8s cluster to try by the way.

What do you think @pearj?

arnaud-deprez:
Hi @diemol,

This is indeed the same command for Kubernetes and Openshift, so I don't plan different documentation for this.
But as I said it might be that setting `securityContext` with `helm template --name zalenium --set hub.securityContext.enabled=true` is required in some vanilla k8s setup to overcome #631 (sorry if this wasn't clear).
As I said too, it works fine without that `securityContext` set on both `minikube` and `minishift` and I can see the dashboard and videos, so I don't know how should I validate against #631 to be 100% sure that the `initContainer` can be removed.

diemol:
Ah ok, thanks, get it. Let's wait for @pearj's opinion.

rodor87:
Would it make sense to add this to the official helm charts repository? I see there's a chart for selenium grid, but none for zalenium

https://github.com/helm/charts/tree/master/stable/selenium

rodor87:
@arnaud-deprez Yes, that's what I mean as well. It makes sense to make the change here first, cut a new release and as part of a new PR migrate to the official helm charts

pearj:
@arnaud-deprez what are you thoughts on supporting the OpenShift `DeploymentConfig`? When using OpenShift I prefer `DeploymentConfig` over `Deployment` especially for development of zalenium.

I think the [template section](https://github.com/zalando/zalenium/blob/7ad0979e9993bc392c07f5400025e8ff70889389/charts/zalenium/templates/deployment.yaml#L13-L160) of `deployment.yaml` can be common between a `DeploymentConfig` and a `Deployment`.  Supporting this would also allow us a `BuildConfig` too, which would make it possible to have one command to setup a build environment for zalenium inside openshift.

Another thing, that I think would be worthwhile is to move all of the `args` to be environment variables instead, like what I did with the openshift template.  For OpenShift at least it is much easier to edit environment variables than it is to edit arguments.
https://github.com/zalando/zalenium/blob/1b065417aff4966e4dc17fed2f1be9cb2a7e7693/docs/k8s/zalenium-template.yaml#L133-L163

In `zalenium.sh` you can see that the environment variables are honoured:
https://github.com/zalando/zalenium/blob/7ad0979e9993bc392c07f5400025e8ff70889389/scripts/zalenium.sh#L3-L26

Another thing that I think is work keeping from the openshift template is using secrets for the grid username and password.  Then inject the secret into the environment variable.

https://github.com/zalando/zalenium/blob/1b065417aff4966e4dc17fed2f1be9cb2a7e7693/docs/k8s/zalenium-template.yaml#L142-L151

See: https://kubernetes.io/docs/concepts/configuration/secret/#using-secrets-as-environment-variables

arnaud-deprez:
Hi @pearj, thank you for your feedback.

I fully agree with you for the environment variables vs args and the secret.

For Openshift deployment vs Kubernetes point, my point was to avoid vendor locking and so support Kubernetes deployment first.
But it should indeed be possible to use `DeploymentConfig` and `BuildConfig` without too much pain.
The main difference I see is the triggers options: 
- optional use of `ImageStreamTag trigger` or not.
- optional `ConfigChange trigger`(if enabled then the behaviour is more or less equivalent to `Deployment` from Kubernetes).

I'll see what I can do and come with a new proposal, but I'm also quite busy right now.

If in the meantime you see other adaptation, please shoot.

arnaud-deprez:
Hi,

Please find the requested changes.

# Remarks

- I put the common PodTemplateSpec section in `_.pod-template.yaml`file. Because it starts with `_`, helm does not generate an output for this.
- I have updated the documentation accordingly.
- I have also tried to refactor the old documentation to use Helm a bit everywhere where it referred previously to `docs/k8s/**`. I didn't find a way to see the preview of this new doc, so if you can please double check this part with me, it will be appreciate :-).

Cheers,

pearj:
Actually I noticed it already used helm in travis, I'll just merge it.

diemol:
Actually, thank you @arnaud-deprez for your time and putting this PR together!

PR #773, opened 2018-11-18
Fixing matching logic 

diemol:
Only waiting for docker-selenium proxies and not for other ones who could match other capabilities.

### DIFF
--- a/src/main/java/de/zalando/ep/zalenium/proxy/AutoStartProxySet.java
+++ b/src/main/java/de/zalando/ep/zalenium/proxy/AutoStartProxySet.java
@@ -17,6 +17,7 @@ import java.util.concurrent.ConcurrentHashMap;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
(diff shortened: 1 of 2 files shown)


src/main/java/de/zalando/ep/zalenium/proxy/AutoStartProxySet.java
```diff
@@ -182,6 +182,16 @@ public void add(RemoteProxy proxy) {
         }
     }
 
+    @Override
+    public void verifyAbilityToHandleDesiredCapabilities(Map<String, Object> desiredCapabilities) {
+        // Using the matcher to see if docker-selenium can handle the desired capabilities. If matches is because the
```
pearj: Is this comment right? If it matches isn’t it because docker-selenium can handle it? Not the cloud proxies?

pearj: Ahh I just read that matcher.  Comment is right.

PR #782, opened 2018-11-23
Saving test information directly to a file instead of keeping it in memory

### LINKED ISSUE #644: Dump logs kept in the memory to the local file during the Runtime
Zalenium Image Version(s):  
3.12.0e
Env: K8S on AWS, deployed with the helm chart from the Zalenium repository.

## Expected Behavior -
Zalenium saves the logs displayed in the Dashboard to the local file so that they are not stored in the memory during the runtime.

## Actual Behavior -
While running a long suite of tests, a lot of RAM is being consumed by the Hub pod. The picture below shows the amount of RAM consumed by the pod with the video recording turned off:
zrzut ekranu 2018-07-06 o 13 21 55

The significant drop in the consumption was observed after using the cleanup functionality in the dashboard. 

Actual behaviour is, quoting @diemol 
> the array is kept in memory and dump to a file when Zalenium shuts down

diemol:
Fixes #644

### DIFF
--- a/src/main/java/de/zalando/ep/zalenium/dashboard/Dashboard.java
+++ b/src/main/java/de/zalando/ep/zalenium/dashboard/Dashboard.java
@@ -5,7 +5,6 @@ import com.google.gson.Gson;
 import com.google.gson.GsonBuilder;
 import com.google.gson.JsonObject;
 import com.google.gson.JsonParser;
-import com.google.gson.reflect.TypeToken;
 import de.zalando.ep.zalenium.util.CommonProxyUtilities;
 import de.zalando.ep.zalenium.util.Environment;
 
@@ -13,12 +12,10 @@ import org.apache.commons.io.FileUtils;
 
 import java.io.File;
 import java.io.IOException;
-import java.lang.reflect.Type;
 import java.util.ArrayList;
 import java.util.Date;
 import java.util.List;
 import java.util.Map;
-import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.stream.Collectors;
 
 import org.slf4j.Logger;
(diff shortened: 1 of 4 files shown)


docker/Dockerfile
```diff
@@ -1,9 +1,24 @@
...
+
+# To overwrite the build args use:
+#  docker build ... --build-arg UBUNTU_DATE=20171006
+ARG UBUNTU_FLAVOR=xenial
+ARG UBUNTU_DATE=20181113
+
+#== Ubuntu xenial is 16.04, i.e. FROM ubuntu:16.04
+# Find latest images at https://hub.docker.com/r/library/ubuntu/
+# Layer size: ~122 MB
+FROM ubuntu:${UBUNTU_FLAVOR}-${UBUNTU_DATE}
```
pearj: I think args in the from value creates issues for building inside openshift, but I can always modify by hand to workaround that

src/main/java/de/zalando/ep/zalenium/dashboard/Dashboard.java
```diff
@@ -250,29 +250,36 @@ public static void synchronizeExecutedTestsValues(File testCountFile) {
...
+    public static void dumpTestInformationToFile(List<TestInformation> testInformationList, boolean replace) {
         try {
-            if (executedTestsInformation.size() > 0) {
-                File testInformationFile = new File(getLocalVideosPath(), TEST_INFORMATION_FILE);
-                Gson gson = new GsonBuilder().setPrettyPrinting().create();
-                FileUtils.writeStringToFile(testInformationFile, gson.toJson(executedTestsInformation), UTF_8);
+            File testInformationFile = new File(getLocalVideosPath(), TEST_INFORMATION_FILE);
+            List<TestInformation> testInformation = new ArrayList<>();
+            if (!replace) {
+                testInformation = loadTestInformationFromFile();
```
pearj: So, it sounds like every 50 tests it will have to load in all of the existing tests to write the file? Doesn't that mean it will still use the same amount of memory? But it will just release it quicker?

How does that replace field work?  Maybe if you could configure it to only keep the last x tests, and then it just pops off the old ones?

I guess if there was a different data structure where you could just append to a file, and you didn't need to keep it in memory then that could make it easier.  Or maybe we've reached the point where we need a lightweight database or something?

docker/Dockerfile
diemol: No worries, I will revert and use the previous format.

src/main/java/de/zalando/ep/zalenium/dashboard/Dashboard.java
diemol: Yes, in the end it is just releasing the memory quicker so the garbage collector frees it sooner. When I wrote it I thought it was a good idea, but I think it just makes things more complicated. Afterwards I came to the same conclusion, it will be easier just to append to a file (I was just being lazy and wanted to keep the formatting through the Gson methods).

I will rework it and implement it in a more simple way.

diemol: Thanks for the review btw.

diemol: I reworked it using the append to a file approach, I am more satisfied with this one than the previous one.
This is the first step I wanted to take to actually render the dashboard from an API call instead of concatenating HTML files in Java. This should also allow people in the near future to use that API call to get the info they need to fetch videos or logs.

PR #797, opened 2018-12-13
Issue-794

ramapalani:
Enable Zalenium UI pages to be served from a non root context

Also change tab in secret.yaml to two spaces

### DIFF
--- a/charts/zalenium/templates/_pod-template.yaml
+++ b/charts/zalenium/templates/_pod-template.yaml
@@ -107,6 +107,8 @@ spec:
           value: {{ .Values.hub.keepOnlyFailedTests | quote }}
         - name: RETENTION_PERIOD
           value: {{ .Values.hub.retentionPeriod | quote }}
+        - name: CONTEXT_PATH
+          value: {{ .Values.ingress.contextpath | quote }}
         {{- if .Values.hub.sauceLabsEnabled }}
         - name: SAUCE_LABS_ENABLED
           value: {{ .Values.hub.sauceLabsEnabled | quote }}
--- a/charts/zalenium/templates/_pod-template.yaml
+++ b/charts/zalenium/templates/_pod-template.yaml
@@ -108,7 +108,7 @@ spec:
         - name: RETENTION_PERIOD
           value: {{ .Values.hub.retentionPeriod | quote }}
         - name: CONTEXT_PATH
-          value: {{ .Values.ingress.contextpath | quote }}
+          value: {{ .Values.ingress.path | quote }}
         {{- if .Values.hub.sauceLabsEnabled }}
         - name: SAUCE_LABS_ENABLED
           value: {{ .Values.hub.sauceLabsEnabled | quote }}
--- a/charts/zalenium/templates/ingress.yaml
+++ b/charts/zalenium/templates/ingress.yaml
@@ -18,7 +18,7 @@ spec:
   - host: {{ .Values.ingress.hostname }}
     http:
       paths:
-      - path: {{ .Values.ingress.contextpath }}
+      - path: {{ .Values.ingress.path }}
         backend:
           serviceName: {{ template "zalenium.fullname" . }}
           servicePort: {{ .Values.hub.port }}
--- a/charts/zalenium/templates/ingress.yaml
+++ b/charts/zalenium/templates/ingress.yaml
@@ -18,7 +18,7 @@ spec:
   - host: {{ .Values.ingress.hostname }}
     http:
       paths:
-      - path: /{{ .Values.ingress.contextpath }}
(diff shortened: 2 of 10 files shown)


charts/zalenium/templates/ingress.yaml
```diff
@@ -18,7 +18,7 @@ spec:
   - host: {{ .Values.ingress.hostname }}
     http:
       paths:
-      - path: /
+      - path: /{{ .Values.ingress.contextpath }}
```
arnaud-deprez: I would prefer to have the / part of the injected value if possible and rename the variable to path so it's a one on one match with the actual ingress configuration:

```suggestion
      - path: {{ .Values.ingress.path }}
```

charts/zalenium/values.yaml
```diff
@@ -143,15 +143,20 @@ persistence:
...
 
 ingress:
-  enabled: false
+  enabled: true
   tls: false
   # secretName: my-tls-cert # only needed if tls above is true
   hostname: zalenium.foobar.com
+  # context path without the forward slash (/)
+  # leaving this empty would run zalenium in the root context (/)
+  contextpath:
```
arnaud-deprez: Then put a default value so it is just easier to enable/disable the ingress: 
```suggestion
  contextpath: /
```

charts/zalenium/values.yaml
```diff
@@ -143,15 +143,20 @@ persistence:
     size: 80Gi
 
 ingress:
-  enabled: false
+  enabled: true
```
arnaud-deprez: Most of the public chart tend to disable it by default as in some cloud provider, it is just easier to play with a service of type `LoadBalancer`. I think it is better to follow this kind of guideline.

ramapalani:
*@arnaud-deprez *thanks for reviewing the code. I'll make these changes and
update the PR.

diemol:
Thanks for the review @arnaud-deprez 👍

charts/zalenium/templates/ingress.yaml
ramapalani: Made this change

charts/zalenium/values.yaml
ramapalani: Made this change

ramapalani:
@arnaud-deprez  Can you review again?

charts/zalenium/values.yaml
```diff
@@ -147,11 +147,15 @@ ingress:
   tls: false
   # secretName: my-tls-cert # only needed if tls above is true
   hostname: zalenium.foobar.com
+  # context path without the forward slash (/)
+  # leaving this empty would run zalenium in the root context (/)
+  contextpath: /
   annotations:
-    # kubernetes.io/ingress.class: "nginx"
+    kubernetes.io/ingress.class: "nginx"
```
arnaud-deprez: please, leave this line commented as no everyone is using nginx ingress.
```suggestion
    # kubernetes.io/ingress.class: "nginx"
```

charts/zalenium/templates/_pod-template.yaml
```diff
@@ -107,6 +107,8 @@ spec:
           value: {{ .Values.hub.keepOnlyFailedTests | quote }}
         - name: RETENTION_PERIOD
           value: {{ .Values.hub.retentionPeriod | quote }}
+        - name: CONTEXT_PATH
+          value: {{ .Values.ingress.contextpath | quote }}
```
arnaud-deprez: Can you please use the same name `path` as the Kubernetes Ingress object. It is just more intuitive IMO.
```suggestion
           value: {{ .Values.ingress.path | quote }}
```

charts/zalenium/templates/ingress.yaml
```diff
@@ -18,7 +18,7 @@ spec:
   - host: {{ .Values.ingress.hostname }}
     http:
       paths:
-      - path: /
+      - path: {{ .Values.ingress.contextpath }}
```
arnaud-deprez: Same remark: use the same name `path` as the Kubernetes Ingress object.

charts/zalenium/values.yaml
```diff
@@ -147,11 +147,15 @@ ingress:
   tls: false
   # secretName: my-tls-cert # only needed if tls above is true
   hostname: zalenium.foobar.com
+  # context path without the forward slash (/)
+  # leaving this empty would run zalenium in the root context (/)
+  contextpath: /
```
arnaud-deprez: Same remark: Can you please use the same name `path` as the Kubernetes Ingress object.

diemol:
Thanks @ramapalani for your contribution! 

Thanks @arnaud-deprez for the review 👍

Issue #794, opened 2018-12-10
Enable Zalenium to run in a non-root context


ramapalani:
In my Kubernetes cluster, I have a ngnix ingress (with a public domain name, say hostname.com). I use this ingress for more than one application.  I want to have the same ingress for Zalenium too.  So I tried exposing Zalenium through a non-root context (say hostname.com/zalenium).  But Zalenium assumes that it would be running only in root context and many pages failed to load. 

Zalenium Image Version(s):  
3
Docker Version:  
17.3.2
If using docker-compose, version:  
NA
OS: 
RHEL
Docker Command to start Zalenium:  
NA
If using Kubernetes, specify your environment, and if relevant your manifests:  
Used helm charts in Zalenium repo with these changes
https://github.com/ramapalani/zalenium/commit/330ace0ce2ec7427d5ad4332774ab5a6084da575?w=1

helm install --name zal1 charts/zalenium

## Expected Behavior -
hostname.com/zalenium/dashboard works good.  I expect other UI pages to work the same.
## Actual Behavior -
This page: /zalenium/grid/admin/live
* couldn't find all static content (css, images,javascript...)
* VNC Iframe results in 404 because it tries to load /vnc/*

When I manually opened hostname.com//zalenium/vnc/host/100.97.233.90/port/50000/?nginx=100.97.233.90:50000&view_only=true
The page loaded but couldn't connect to VNC because, websockify was directly pointing to /proxy

It would be great to fix this, let me know how I could help this fixed

ramapalani:
Not working UI pages
![live](https://user-images.githubusercontent.com/7064234/49760059-4b09ab00-fc78-11e8-8f3b-1f83e351c8c8.png)
![vnc](https://user-images.githubusercontent.com/7064234/49760069-52c94f80-fc78-11e8-8e93-b6d7ceec8d62.png)

Working UI page
![dashboard](https://user-images.githubusercontent.com/7064234/49760087-5f4da800-fc78-11e8-8272-e512c0977070.png)

diemol:
Hi @ramapalani,

Yes, the dashboard comes through an nginx config and the other ones are rendered by Jetty. These are the files for the LiveView
https://github.com/zalando/zalenium/blob/master/src/main/java/de/zalando/ep/zalenium/servlet/LivePreviewServlet.java
https://github.com/zalando/zalenium/blob/master/src/main/resources/html_templates/live_preview_servlet.html

Maybe you can give us a hand to make this work for your use case as well?

[...]

PR #811, opened 2018-12-29
Issue:794 Get NoVNC web socket to work from LivePreview (/grid/admin/live)

ramapalani:
This PR addresses the last part of https://github.com/zalando/zalenium/issues/794 (getting a noVNC work for
non root
context)
* LiveNodeHtmlRenderer: Pass context path to NoVNC web socket
(/contextpath/proxy/seleniumnode:novncport/websockify)
* Values.yaml: Delete the wrong comment for path.
* .gitignore: Ignore IDE specific directories

### DIFF
--- a/charts/zalenium/values.yaml
+++ b/charts/zalenium/values.yaml
@@ -1,10 +1,10 @@
 hub:
   ## The repository and image
-  ## ref: https://hub.docker.com/r/selenium/hub/
+  ## ref: https://hub.docker.com/r/dosel/zalenium
   image: "dosel/zalenium"
 
   ## The tag for the image
-  ## ref: https://hub.docker.com/r/selenium/hub/tags/
+  ## ref: https://hub.docker.com/r/dosel/zalenium/tags
   tag: "3"
 
   ## Specify a imagePullPolicy
@@ -16,23 +16,6 @@ hub:
 
   ## Timeout for probe Hub readiness via HTTP request on Hub console
   readinessTimeout: 1
-
-  ## Set the JAVA_OPTS environment variable
-  ## If you find your selenium hub is OOMKilled, try adding -XX:+UseSerialGC
-  javaOpts: "-Xmx400m -XX:+UseSerialGC"
-
-  ## Set the SE_OPTS environment variable
-  ## ref: http://www.seleniumhq.org/docs/07_selenium_grid.jsp#node-configuration
-  # seOpts:
-
-  ## Defining a JMX port will open the port on the container, however, it
-  ## requires additional javaOpts, ie
-  ## javaOpts: >
-  ##   -Dcom.sun.management.jmxremote.port=4000
-  ##   -Dcom.sun.management.jmxremote.authenticate=false
-  ##   -Dcom.sun.management.jmxremote.ssl=false
-  ## ref: http://openjdk.java.net/groups/jmx/
-  # jmxPort: 4000
 
   ## Pod Security Context
   ## ref: https://kubernetes.io/docs/tasks/configure-pod-container/security-context/
--- a/charts/zalenium/values.yaml
+++ b/charts/zalenium/values.yaml
@@ -147,8 +147,6 @@ ingress:
   tls: false
   # secretName: my-tls-cert # only needed if tls above is true
   hostname: zalenium.foobar.com
-  # context path without the forward slash (/)
-  # leaving this empty would run zalenium in the root context (/)
   path: /
   annotations:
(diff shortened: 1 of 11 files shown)


charts/zalenium/values.yaml
```diff
@@ -147,8 +147,6 @@ ingress:
   tls: false
   # secretName: my-tls-cert # only needed if tls above is true
   hostname: zalenium.foobar.com
-  # context path without the forward slash (/)
-  # leaving this empty would run zalenium in the root context (/)
```
diemol: Why are you removing the comments? Perhaps it would be nice to leave something more concrete there explaining what `path` is used for,

src/main/java/de/zalando/ep/zalenium/servlet/renderer/LiveNodeHtmlRenderer.java
```diff
@@ -55,9 +55,15 @@ public String renderSummary() {
         // Adding live preview
         int noVncPort = proxy.getRegistration().getNoVncPort();
         String noVncIpAddress = proxy.getRegistration().getIpAddress();
-        String noVncViewBaseUrl = "%s/vnc/host/%s/port/%s/?nginx=%s:%s&view_only=%s";
-        String noVncReadOnlyUrl = String.format(noVncViewBaseUrl, contextPath, noVncIpAddress, noVncPort, noVncIpAddress, noVncPort, "true");
-        String noVncInteractUrl = String.format(noVncViewBaseUrl, contextPath, noVncIpAddress, noVncPort, noVncIpAddress, noVncPort, "false");
+        String websockifyContextPath = null;
+        if (contextPath == null || contextPath.isEmpty()) {
```
diemol: For this `if`, you could just use `Strings.isNullOrEmpty()`.

Or even leave the whole thing like:
```suggestion
        String websockifyContextPath = Strings.isNullOrEmpty(contextPath) ? "" : contextPath + "/";
```

charts/zalenium/values.yaml
ramapalani: added appropriate comments.

ramapalani:
@diemol  Updated the PR based on your review comments

pearj:
How does this work with the embedded nginx? I can see you haven't changed the [nginx.conf](https://github.com/ramapalani/zalenium/blob/9e531dd75caf0e86e4b1ee0c4b2cf54472319ca5/docker/nginx.conf#L64).

Does your kubernetes ingress server do url rewriting? So that calls to `/yourprefix/vnc/host/machine1/port/40000` end up hitting the backend with `/vnc/host/machine1/port/40000`?

So your change is simply about change how links are generated, rather than completely changing the root context?

It's worth noting here that the OpenShift Router can't do URL rewriting, so from what I can see so far, these changes don't go far enough to make it work in that context.

From what I gather if you're using the kubernetes nginx ingress, you can do URL rewriting via a specific nginx annotation, so it's not portable and wouldn't even work in [GKE for example](https://github.com/kubernetes/ingress-gce/issues/109).

ramapalani:
@pearj 
My requirement is to have same kubernetes ingress host name (say hostname.com) and have multiple applications behind it.  So I want Zalenium to listen in a different subcontext (say hostname.com/zalenium). 

I'm using ingress-nginx as my kubernetes ingress controller and Yes I'm relying on ingress-nginx rewrite url annotation.  I wanted to keep my changes to minimal, so all I did is to generate links in html with a context root (if one is provided)

If zalenium should be listening on a context (for all kubernetes ingress controllers) then all location entries in this [nginx.conf](https://github.com/zalando/zalenium/blob/9e531dd75caf0e86e4b1ee0c4b2cf54472319ca5/docker/nginx.conf)   should be updated with the provided context.  Shall I do this?  Or do you have an alternate solution?

ramapalani:
@pearj Could you review my last update.  I moved the rewrite into zalenium's nginx server, so that it could work across all ingress controllers.

ramapalani:
@diemol or @pearj Could you review the latest changes?

docker/nginx.conf
```diff
@@ -24,15 +24,18 @@ http {
         }
         auth_basic $auth_basic;
 
+        # rewrite {{contextPath}}/(.*) /$1 break;
+        # rewrite {{contextPath}} / break;
+
         location / {
             proxy_pass http://127.0.0.1:4445;
         }
+        # location /dashboard {
```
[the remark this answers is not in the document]
ramapalani: I made a change and it works now

diemol:
Thanks @ramapalani! Looks good to me.

PR #832, opened 2019-01-15
fix rbac: pod exec verb does not exist

arnaud-deprez:
### Description
There is an issue in the rbac: verb `exec` does not exist for resource `pod`.
I've replaced it by deleteCollection (in case we need it).

### Motivation and Context
The consequence of this issue is that a user that is not cluster-admin will be rejected when trying to deploy zalenium because it is giving higher privileges than himself to service account (behaviour observed on Openshift).

### How Has This Been Tested?
I've tested it manually on Openshift

### Types of changes

### Checklist:


arnaud-deprez:
I don't know actually, I didn't check the code neither.
I put this configuration because it existed in the other vanilla kubernetes/openshift configuration, but it should indeed be nice to review the code and list the required privileges.

But then I would need more time unless someone else with better knowledge of the code base can do it quickly.

Cheers,

diemol:
I remember that in some unit tests I had to mock the calls to `exec` for pods, but I am not sure if that is not needed anymore.
Let's wait for @pearj's comment.

pearj:
Ahh right, I think `exec` is not needed as it was a misunderstanding of how rbac is setup in regards to executing commands on pods. I thought that you would do an `exec` on a `pod` resource.  But instead, it looks like you do a `create` on a `pod/exec` resource.

I would have only tested with cluster-admin, so that's why I didn't see this issue.

FYI the reason it needs to execute a command on a pod is when video recordings are started/stopped and for copying the logs and videos from a pod after a test finishes.

Issue #847, opened 2019-01-22
Cannot see videos in dashboard


SilverXXX:
## 🐛 Bug Report

When accessing the dashboard, a list of video is shown but there is no playback, and many logs with stacktraces are written

## To Reproduce

Install with :
helm install --set hub.maxDockerSeleniumContainers=100 --set hub.desiredContainers=10 --set ingress.enabled=true --name zalenium-name charts\zalenium\
then launching any test

## Expected behavior

Videos should be playing inside the dashboard

## Actual behavior

Videos are not showing at all, and these lines appears inside pod logs:
13:29:38.402 [http://10.40.0.40:40000] WARN  d.z.e.z.p.DockerSeleniumRemoteProxy - Error while copying the video
java.nio.file.NoSuchFileException: /home/seluser/videos/build_2019_01_22T13_28_15_280Z/zalenium_webRtcOperator_1_firefox_LINUX_20190122132832642_FAILED.mp4
	at sun.nio.fs.UnixException.translateToIOException(UnixException.java:86)
	at sun.nio.fs.UnixException.rethrowAsIOException(UnixException.java:102)
	at sun.nio.fs.UnixException.rethrowAsIOException(UnixException.java:107)
	at sun.nio.fs.UnixFileSystemProvider.newByteChannel(UnixFileSystemProvider.java:214)
	at java.nio.file.spi.FileSystemProvider.newOutputStream(FileSystemProvider.java:434)
	at java.nio.file.Files.newOutputStream(Files.java:216)
	at java.nio.file.Files.copy(Files.java:3016)
	at de.zalando.ep.zalenium.proxy.DockerSeleniumRemoteProxy.copyVideos(DockerSeleniumRemoteProxy.java:639)
	at de.zalando.ep.zalenium.proxy.DockerSeleniumRemoteProxy.processContainerAction(DockerSeleniumRemoteProxy.java:609)
	at de.zalando.ep.zalenium.proxy.DockerSeleniumRemoteProxy.processContainerAction(DockerSeleniumRemoteProxy.java:598)
	at de.zalando.ep.zalenium.proxy.DockerSeleniumRemoteProxy.videoRecording(DockerSeleniumRemoteProxy.java:562)
	at de.zalando.ep.zalenium.proxy.DockerSeleniumRemoteProxy.cleanupNode(DockerSeleniumRemoteProxy.java:749)
[...]

PR #851, opened 2019-01-29
Fix issue #847

pearj:
Fixes issue #847 

### Description
* Add buildName to the videosFolder instead of the file name
* Ensure that videos in build directories are removed

### How Has This Been Tested?
Tested by hand in OpenShift

### Types of changes


diemol:
Thanks @pearj!

PR #852, opened 2019-01-29
Dashboard Reworked

### LINKED ISSUE #142: Videos dashboard leaks memory
I've noticed that if you leave the videos dashboard open for a long time it leaks a lot of memory to the point where Google Chrome kills it for using too much memory.

When I first open it the Google Chrome task manager says it's using about 180MB, then after an hour it was up to 700MB, and that's without any new tests running through Zalenium.

diemol:
Moving from the concept of concatenating Strings into forming an HTML file to use a simple and maybe suboptimal endpoint to retrieve the test information and display it in a nicer way.

dashboard

Fixes #142


diemol:
I also wanted some JS framework but still wanted to keep the offline dashboard working, and I didn't know how to achieve that with a framework.
But yes, eventually we should do something like that :)

PR #854, opened 2019-01-31
Add support to use imagePullSecret on pod creation

leslau:
Make it possible to use imagePullSecret to pull zalenium images from private repositories.

Pods created by zalenium will use the same imagePullSecret.

### DIFF
--- a/charts/zalenium/templates/_pod-template.yaml
+++ b/charts/zalenium/templates/_pod-template.yaml
@@ -9,6 +9,10 @@ spec:
     fsGroup: {{ .Values.hub.securityContext.fsGroup }}
     runAsUser: {{ .Values.hub.securityContext.runAsUser }}
   {{- end }}
+  {{- if .Values.hub.pullSecrets }}
+  imagePullSecrets:
+    - name: {{ .Values.hub.imagePullSecret }}
+  {{- end}}
   volumes:
     - name: {{ template "zalenium.fullname" . }}-videos
     {{- if .Values.persistence.video.enabled }}
--- a/charts/zalenium/values.yaml
+++ b/charts/zalenium/values.yaml
@@ -11,6 +11,10 @@ hub:
   ## ref: http://kubernetes.io/docs/user-guide/images/#pre-pulling-images
   pullPolicy: "IfNotPresent"
 
+  ## Specify secrets to pull images from private repositories
+  ## ref: https://kubernetes.io/docs/concepts/containers/images/#specifying-imagepullsecrets-on-a-pod
+  #imagePullSecret: "secret1"
+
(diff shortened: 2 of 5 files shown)


charts/zalenium/templates/_pod-template.yaml
```diff
@@ -9,6 +9,10 @@ spec:
     fsGroup: {{ .Values.hub.securityContext.fsGroup }}
     runAsUser: {{ .Values.hub.securityContext.runAsUser }}
   {{- end }}
+  {{- if .Values.hub.pullSecrets }}
+  imagePullSecrets:
+    - name: {{ .Values.hub.pullSecrets }}
```
arnaud-deprez: Please rename `pullSecrets` to `imagePullSecret` to follow kubernetes pod template semantic and as you set only one secret.

charts/zalenium/values.yaml
```diff
@@ -11,6 +11,10 @@ hub:
   ## ref: http://kubernetes.io/docs/user-guide/images/#pre-pulling-images
   pullPolicy: "IfNotPresent"
 
+  ## Specify secrets to pull images from private repositories
+  ## ref: https://kubernetes.io/docs/concepts/containers/images/#specifying-imagepullsecrets-on-a-pod
+  #pullSecrets: "secret1"
```
arnaud-deprez: Same: imagePullSecret

src/main/java/de/zalando/ep/zalenium/container/kubernetes/KubernetesContainerClient.java
```diff
@@ -102,6 +103,7 @@ public KubernetesContainerClient(Environment environment,
             discoverHostAliases();
             discoverNodeSelector();
             discoverTolerations();
+            discoverPullSecrets();
```
arnaud-deprez: I think it's better to keep the full name as well. Please rename to `discoverImagePullSecrets()`

src/test/java/de/zalando/ep/zalenium/container/kubernetes/PodConfigurationTest.java
```diff
@@ -128,4 +128,14 @@ public void setTolerations() {
         assertThat(podConfiguration.getTolerations().size(), is(1));
         assertThat(podConfiguration.getTolerations().get(0), is(toleration));
     }
+
+    @Test
+    public void setPullSecrets() {
```
arnaud-deprez: I would prefer to rename it as `testSetImagePullSecrets` as well

diemol:
Thanks for your contribution @leslau and @arnaud-deprez for the review!

PR #875, opened 2019-02-20
fix role.yaml

arnaud-deprez:
### Description
Rename verb deleteCollection to deletecollection according to the [doc](https://kubernetes.io/docs/reference/access-authn-authz/authorization/#determine-the-request-verb)

### Motivation and Context
If a user a non cluster-admin user tries to deploy zalenium in a project, it will fails because `deleteCollection` does not match the permission `deletecollection`.
This is a summary of the issue, the error message just state that it tries to create a role with higher privilege than the user has.

### How Has This Been Tested?
Tested on Openshift v3.10

### Types of changes

### Checklist:

### DIFF
--- a/charts/zalenium/templates/role.yaml
+++ b/charts/zalenium/templates/role.yaml
@@ -14,7 +14,7 @@ rules:
   verbs:
   - create
   - delete
-  - deleteCollection
+  - deletecollection
   - get
   - list
   - watch


diemol:
Thanks for your contribution @arnaud-deprez!

arnaud-deprez:
With pleasures @diemol ;-)

PR #888, opened 2019-03-06
Feature/zalenium helm repo

axozoid:
### Issue statement
In order to install this helm chart people need to have it in the local helm repo (e.g., download/clone this github repo) which might be not very convenient. 

### Description

1. Updated `appVersion` to point to the version of `docker-selenium` image (`3.141.59`);
2. Added files necessary for helm to be able to use this GitHub repo as a helm repo;
3. Updated the documentation on how to use this new feature;

### Motivation and Context
Having a possibility to install the helm chart straight from this GItHub repo would be beneficial and convenient.

### How Has This Been Tested?
* After implementing these changes I was able to deploy Zalenium using helm, without a need to clone this repo. 
* Tested in Kubernetes v1.10.11 running in AWS.
* Changes do not affect any areas of the code.

### Types of changes

### Checklist:


axozoid:
I've no idea why its failing on 'Step 11/49'. I can build the image locally without any issues.

charts/zalenium/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 name: zalenium
 version: 0.0.1
-appVersion: 3.4.0
+appVersion: 3.141.59
```
arnaud-deprez: should be aligned with version in index.yaml

charts/zalenium/index.yaml
```diff
@@ -0,0 +1,21 @@
+apiVersion: v1
+entries:
+  zalenium:
+  - appVersion: 3.4.0
```
arnaud-deprez: aligned with 3.141.59

charts/zalenium/index.yaml
```diff
@@ -0,0 +1,21 @@
...
+    - qa
+    maintainers:
+    - email: diemol@gmail.com
+      name: diemol
+    name: zalenium
+    sources:
+    - https://github.com/zalando/zalenium
+    urls:
+    - zalenium-0.0.1.tgz
+    version: 0.0.1
```
arnaud-deprez: aligned with 3.141.59

axozoid:
Thank you @arnaud-deprez,
I'm happy to follow whatever pattern you guys think is the best.

Regards.

diemol:
Hi @axozoid, thanks for your PR, makes completely sense, instead of cloning the repo to use the chart, to do something like this.

I completely agree with @arnaud-deprez (by the way, thanks for the review), this should be part of an automated release process, where it gets built in Travis and automatically added to the repo.

Right now there is a small manual part that needs to be done for every release, which is me generating the tag and pushing it from a release branch, this because there are some restrictions under the org. Nevertheless, this will change soon, more details to come.

For now, it would be nice to follow @arnaud-deprez's advice, have the version number in one place, follow the version rules of helm, and generate the `.tgz` file as part of the release.

arnaud-deprez:
Hi @diemol,

In order to be able to apply the template from the git repo, we need to publish an `index.yaml` as well and it should reflect the server structure that helm expect.

So to summarise, while publishing the `.tgz` will help us to download, untar and then apply helm chart, it is still not straightforward.

Instead, what other people are doing, they are publishing the `.tgz` and the `index.yaml` in github pages.

Knowing that, I'm wondering how it is managed behind this https://opensource.zalando.com/.
Is https://opensource.zalando.com/zalenium simply a redirection to this github page, or is this website hosted somewhere else using other techniques ?

IMO, it should be nice to use something like: 

```sh
# global charts for zalando
helm repo add zalando https://opensource.zalando.com/charts
# if they use tiller
helm install zalando/zalenium
# or specific to zalenium
helm repo add zalenium https://opensource.zalando.com/zalenium/charts
# if they use tiller
helm install zalenium/zalenium
```

diemol:
@arnaud-deprez yeah, it would be nice to have it the way you propose. 
I am not working in Zalando anymore, but I can imagine that it is just a redirect to the GitHub pages.

Note aside, which is kind of relevant to this, the repo will be moved to another org in 1-2 months, where we can have a bit more freedom to do what you suggest. 

Having said that, what do you think we should have in this PR for now?

charts/zalenium/index.yaml
axozoid: Is now aligned.

charts/zalenium/Chart.yaml
axozoid: Done.

arnaud-deprez:
Hi @diemol,

I think we can still publish this to the github pages of the current repo.
Even if the proxy is not a simple redirection we could use the github url directly.

I will need some time, unless @axozoid can look into it.

arnaud-deprez:
Agree.
I'm just missing some time right now. So I can't garantee to do it this week.
@axozoid do you feel confident to make it or do you prefer I handle it ?

arnaud-deprez:
Hi @diemol,

In order to publish static content on github pages, we need to enable this in a branch in the project settings.
So basically, if you have access to it, we can make it, otherwise I'm afraid we have to wait or find another place to publish our helm charts.

diemol:
@arnaud-deprez yes, everything in the docs directory is part of GitHub pages, I did that setup before leaving Zalando, so if we put something there, it will be available,

diemol:
I think I will merge this now, to avoid keeping it hanging. Additionally I will create an issue to automate the helm chart generation, since we don't have 100% control of the repo now, it is hard to do it right now. This should change in the upcoming weeks. See #917

PR #890, opened 2019-03-07
Use the same network for both hub and nodes.

JorisVanEijden:
### Description
Expose the hub IP address that is communicated to the nodes in DockeredSeleniumStarter.
In DockerContainerClient use this IP address to determine the correct network to use for creating the nodes.  

### Motivation and Context
Zalenium puts the nodes in the first network it finds on the hub that isn't "bridge" but it tells the nodes to contact it on the first non-loopback interface it finds. These are not necessarily the same network.
This causes the nodes to be unable to connect to the hub to register themselves.

### How Has This Been Tested?
I used Lando (https://github.com/lando/lando) which run a docker-compose setup with multiple networks. Adding zalenium here leads to the issue described in the context.
.lando.yml:
```yaml
name: myproject

recipe: drupal7

config:
  webroot: web
  php: 7.1
  via: apache:2.4
  database: postgres

services:
  zalenium:
    type: compose
    services:
      image: dosel/zalenium
      ports:
        - "4444:4444"
      command: ["entry.sh", "start"]
      volumes:
        - /var/run/docker.sock:/var/run/docker.sock
        - /tmp/mounted:/tmp/mounted
        - /dev/shm:/dev/shm
      environment:
        - PULL_SELENIUM_IMAGE=true
        - SEND_ANONYMOUS_USAGE_INFO=false
        - DEBUG_ENABLED=true
    portforward: true

proxy:
  zalenium:
    - zalenium.lndo.site:4444
```
In this case the hub has 3 networks. Zalenium picks one of them to create the nodes on. It picks a different one to get it's IP address from to communicate to the nodes (SELENIUM_HUB).

Then I did the same with a new zalenium image built from the branch with my changes and the problem went away.
The IP communicated to the nodes is on the same network as the nodes are created on.

### Types of changes

### Checklist:

### DIFF
--- a/src/main/java/de/zalando/ep/zalenium/container/DockerContainerClient.java
+++ b/src/main/java/de/zalando/ep/zalenium/container/DockerContainerClient.java
@@ -595,10 +595,11 @@ public class DockerContainerClient implements ContainerClient {
                     "variable ZALENIUM_CONTAINER_NAME has an appropriate value", zaleniumContainerName));
         }
         try {
+            String hubIpAddress = DockeredSeleniumStarter.getHubIpAddress();
             ContainerInfo containerInfo = dockerClient.inspectContainer(zaleniumContainerId);
             ImmutableMap<String, AttachedNetwork> networks = containerInfo.networkSettings().networks();
             for (Map.Entry<String, AttachedNetwork> networkEntry : networks.entrySet()) {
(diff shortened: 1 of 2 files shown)


diemol:
Hi @JorisVanEijden, thanks for taking the time to send a PR!
After checking the code, things _shoud_ work. Just let me merge this into a protected branch so we can run all the tests, and then we get a better feeling of it.

PR #892, opened 2019-03-10
Use the same network for both hub and nodes.

diemol:
Running all CI tests for #890


diemol:
Looks good @JorisVanEijden, thanks!

PR #915, opened 2019-03-29
Add log fetching support for BrowserStack

kallekrantz:
### Description
This commit implements BrowserStack support for Selenium and browser driver logs. It does this
by extending the existing implementation to also encapsulate optional authentication for file downloads.

It does not modify existing implementations, in order to keep the blast-radius of the commit as low as possible.

### Motivation and Context
As we're using a limited BrowserStack plan, with only a few users being able to be active. Our develers requested the Selenium and driver logs to be present. This feature tries to add that feature with as small a blast-radius as possible.

### How Has This Been Tested?
This was tested locally using our BrowserStack credentials, I've done my best to try it with several different browsers and devices. If there is a test suite I can run (other than the unit tests) to do more rigorous testing, please let me know!

In terms of impact to other sides of the code, these should be few. The one "concern" I have is the `Accept: */*` I added for in order for log fetching to work. I'm guessing this will have a low impact, but if there we feel like it, I have no problem adding a hook for adding custom Headers for particular file downloads. 

### Types of changes

### Checklist:

- [ ] My code follows the code style of this project. *First Java I've written in a while, I don't know if there's a check I can run locally to check for the code style*
- [ ] My change requires a change to the documentation. *To both this one and the one below, I'm unsure if the lack of log fetching is mentioned anywhere in the documentation, or only in usage of Zalenium*
- [ ] I have added tests to cover my changes. *Unsure, I've modified existing tests, and since this particular code path is already tested, it didn't look crazy to leave as is. Do let me know if you want some tests! More than happy to provide!*

### DIFF
--- a/src/main/java/de/zalando/ep/zalenium/dashboard/TestInformation.java
+++ b/src/main/java/de/zalando/ep/zalenium/dashboard/TestInformation.java
@@ -3,6 +3,8 @@ package de.zalando.ep.zalenium.dashboard;
 
 import com.google.common.base.Strings;
 import com.google.gson.JsonObject;
+
+import de.zalando.ep.zalenium.proxy.RemoteLogFile;
 import de.zalando.ep.zalenium.util.CommonProxyUtilities;
 
 import java.util.Date;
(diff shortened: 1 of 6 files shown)


diemol:
@kallekrantz I am sorry, I squash merged the commits, and somehow I removed the original you as the original author. I will try to rewrite the git history to see if I can re-establish that and give you proper credit for the commit.

PR #924, opened 2019-04-04
Feature openshift route

torstenwalter:
quickly.**

### Description
I added the possibility to create an OpenShift route. TLS is enabled by default, but it can be disabled.
The hostname is optional as OpenShift creates a default for it based on the route's name and the current namespace.

### Motivation and Context

I want to use this chart to deploy Zalenium to OpenShift. I could use Ingress for that, but using routes feels more natural as I can rely on OpenShift to generate a unique host name.

### How Has This Been Tested?

Tested it locally with Minishift, with and without TLS / hostname.

### Types of changes

### Checklist:

- [x] I have updated the documentation accordingly. (README is updated)

I did not find any helm specifc tests to update.

### DIFF
--- a/charts/zalenium/README.md
+++ b/charts/zalenium/README.md
@@ -114,6 +114,10 @@ See Zalenium's [usage examples](https://github.com/zalando/zalenium/blob/master/
 | `hub.basicAuth.enabled` | Enables basic authentication | false |
 | `hub.basicAuth.username` | Username for basic authentication | zalenium |
 | `hub.basicAuth.password` | Password for basic authentication | password |
+| `hub.openshift.route.enabled`  | Set to true if you want to create a route for zalenium | false |
+| `hub.openshift.route.hostname` | If you want to have a specific hostname specify it here | blank |
+| `hub.openshift.route.tls` | Configures tls settings for OpenShift route. Set it to empty if you don't want it | edge termination + redirect |
+
 
 Specify each parameter using the `--set key=value[,key=value]` argument to `helm install`. For example,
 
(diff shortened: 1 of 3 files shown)


arnaud-deprez:
Thanks for your PR @torstenwalter.

From Openshift 3.10, if you create an ingress, it creates a corresponding openshift route.
So I initially think that it didn't make sense to have Openshift route support.

The only facility Openshift provides with its routes is to be able to reuse the same wildcard certificates for all routes. So I guess it make sense to have it for such scenario.

torstenwalter:
@arnaud-deprez You are right! It's basically just to be able to reuse the wildcard certificate. Doest it make sense for that use case?

arnaud-deprez:
Yeah I don't see any objection for it.
However I'm wondering if we should not avoid `tls` by default to have the same behaviour than the usual ingress which is by default to `ingress.tls: false`

torstenwalter:
In my experience it's better to have by default a secure option and to allow users to reduce security if they don't need it.
It's very similar as enabling tbac by default which is documented as helm [best practice](https://helm.sh/docs/chart_best_practices/#rbac-resources-should-be-created-by-default).

As tls certificated are available in most OpenShift installations TLS by default should not cause problems.

arnaud-deprez:
Ok but most default certificate are self-signed by default which might cause some complication on your test side.
If you are using a dev cluster for your test, then most probably your cluster is not exposed to the internet and so you don't really care if it's http or https.

torstenwalter:
I see your point. Although personally I'd like to have everything tls secured even internal communication, but I am fine in chaning this. 

Chart is updated.

arnaud-deprez:
It's ok for me @diemol

PR #949, opened 2019-04-24
Syncing local repo with the upstream.

axozoid:
syncing with the upstream.


axozoid:
GitHub offered to create this PR since my branch was ahead..

PR #950, opened 2019-04-24
Introducing a dedicated port for WebUI

axozoid:
### Description
These changes will add more flexibility for ports' configuration.

### Motivation and Context
Before, there was only one `port` (a hub was listening on) value available for changing, but the problem is that this value was used for both keys in kubernetes service definition `port` and `targetPort` which not always is desired to be the same.
In a case you have some firewalls in place which allow only certain ports to be accessed - you're in trouble. Also, as it turns out, port 4444 is not available over the internet due to being blocked by many telcos.

### How Has This Been Tested?
Tested within our local dev cluster hosted in AWS, using Helm.
In case of serviceType=LoadBalancer, an ELB will be listening on `svcPort`.

### Types of changes

### Checklist:

### DIFF
--- a/charts/zalenium/templates/service.yaml
+++ b/charts/zalenium/templates/service.yaml
@@ -16,7 +16,7 @@ spec:
   sessionAffinity: {{ .Values.hub.serviceSessionAffinity | quote }}
   ports:
   - name: hub
-    port: {{ .Values.hub.port }}
+    port: {{ .Values.hub.svcPort }}
     targetPort: {{ .Values.hub.port }}
   selector:
     app.kubernetes.io/name: {{ include "zalenium.name" . }}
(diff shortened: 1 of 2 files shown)


diemol:
@axozoid I am completely out of context, probably my K8S knowledge is not enough to grasp the idea of this change. Could you please elaborate a bit more?
@davidcollom @arnaud-deprez what do you think?

davidcollom:
This makes sense, as the currently helm chart only allows a 1:1 mapping on ports (I.E: a loadbalancer would listen on 4444 and forward to 4444.)

This change would allow users to expose port 80 outside of the cluster (NodePort/LoadBalancer) but route back to port 4444 for the zalenium-hub.

diemol:
Got it, thanks for the review @davidcollom!

diemol:
Thanks for your contribution @axozoid

axozoid:
My pleasure!

PR #907, opened 2019-03-26
Docker Swarm Implementation

yosserO:
### Description

How to use it:

* Initialize a swarm by running: `docker swarm init`
* Add **at least** one worker to the swarm.
* Deploy a new Stack by running: `docker stack deploy -c <path-to-docker-compose-file> STACK_NAME`.
    - We added an example docker-compose file under _zalenium/docs/docker/docker-compose-swarm.yml_.
    - An attachable  overlay network must be declared in the docker-compose file. This overlay network must be passed to zalenium as the configuration parameter "--swarmOverlayNetwork". In our example we are declaring the overlay network "zalenium". We named our stack "tau". So the corresponding value for the configuration parameter "--swarmOverlayNetwork" is "tau_zalenium".
* We have two different **constraints** for the hub and its nodes:
    - The hub has the constraint `node.role == manager` and will only be deployed on the manager. (Take a look at the example docker-compose file).
    - The nodes have the constraint `node.role==worker`. This constraint is currently hard coded in the _SwarmContainerClient.java_ Class.
* Logs and Videos are currently not supported.
* In order to execute commands on workers we are using this image `datagridsys/skopos-plugin-swarm-exec:latest`. 
* We tested the solution in our own environment but we will be thankful if you test it for your own and give us feedback.
* We update the documention later after positive feedback.
 
### Motivation and Context

Adapt zalenium to be able to run in a docker swarm. 

### How Has This Been Tested?

We tested it manually. For more details please check the **Description**.

### Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [x] Breaking change (fix or feature that would cause existing functionality to change)

### Checklist:

- [ ] My code follows the code style of this project.
- [x] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [x] I have read the **CONTRIBUTING** document.
- [ ] I have added tests to cover my changes.
- [x] All new and existing tests passed.

PR #982, opened 2019-06-02
Docker Swarm Implementation

diemol:
Comes from #907, thanks to @yosserO and @tstern 
Running all CI

### DIFF
--- a/src/main/java/de/zalando/ep/zalenium/container/swarm/SwarmContainerClient.java
+++ b/src/main/java/de/zalando/ep/zalenium/container/swarm/SwarmContainerClient.java
@@ -1,0 +1,655 @@
+package de.zalando.ep.zalenium.container.swarm;
+
+import com.google.common.annotations.VisibleForTesting;
+import com.google.common.collect.ImmutableMap;
+import com.spotify.docker.client.AnsiProgressHandler;
+import com.spotify.docker.client.DefaultDockerClient;
+import com.spotify.docker.client.DockerClient;
+import com.spotify.docker.client.exceptions.DockerException;
+import com.spotify.docker.client.messages.*;
+import com.spotify.docker.client.messages.mount.Mount;
+import com.spotify.docker.client.messages.swarm.*;
+import de.zalando.ep.zalenium.container.ContainerClient;
+import de.zalando.ep.zalenium.container.ContainerClientRegistration;
+import de.zalando.ep.zalenium.container.ContainerCreationStatus;
+import de.zalando.ep.zalenium.proxy.DockeredSeleniumStarter;
+import de.zalando.ep.zalenium.util.Environment;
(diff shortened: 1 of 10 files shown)


diemol:
@yosserO @tstern 
Things look good to me, I am just missing documentation for end users. What we can do is to merge this to master, release the feature as `beta`, and in a subsequent PR you could add documentation.

I'll also appreciatte if you can hangout on the #zalenium channel in the Selenium Slack, in case people have questions.

What do you think?

yosserO:
Hi Diego,
Thank you for reviewing the PR. 
We will add the documentation next Week. We are already in the zalenium and Selenium Slack Channels. If there are questions related to docker swarm, please just mention us in the comment so we can get a notification.

trinhpham:
Really eager to be one of the first users :)
My question is: when and where can I get the beta build?

src/main/java/de/zalando/ep/zalenium/container/swarm/SwarmContainerClient.java
```diff
@@ -0,0 +1,477 @@
...
+                .collect(Collectors.toList());
+        final ContainerSpec containerSpec = buildContainerSpec(flattenedEnvVars, image);
+
+        final TaskSpec taskSpec = buildTaskSpec(containerSpec);
+
+        String noVncPort = envVars.get("NOVNC_PORT");
+        final ServiceSpec serviceSpec = buildServiceSpec(taskSpec, nodePort, noVncPort);
+
+        try {
+            ServiceCreateResponse service = SwarmUtilities.createService(serviceSpec);
```
trinhpham: I think it should be better if we pre-define the worker services, then update theirs scaling like `https://docs.docker.com/engine/swarm/swarm-tutorial/scale-service/`

yosserO: Hi @trinhpham,

Service scaling was our first Approach for creating and deleting Browser tasks. 

The problem was: When scaling down the number of replicas, it doesn't took in account which task is  not active. It killed containers that were actually running tests and made the tests fail. We found no way to specify the task to shutdown, when scaling down the number of replicas. 

https://github.com/moby/moby/issues/33426

If we scale the number of replicas up instead of creating new services. We will have to scale it down when the tests end. And this will lead to a wrong behaviour.

trinhpham: Got your idea. Thanks for explaining it to me :)

diemol:
Ok, let's merge this and do a release!

yosserO:
> Really eager to be one of the first users :)
> My question is: when and where can I get the beta build?

We created a pull request to add documentation: https://github.com/zalando/zalenium/pull/984

Issue #904, opened 2019-03-24
kubernetes securtiyContext is not implemented in node spawning (rbac enabled)


EyalZilberblum:
## 🐛 Bug Report

KubernetesContainerClient not implementing securityContext, which leading to "Error: container has runAsNonRoot and image has non-numeric user (seluser), cannot verify user is non-root
"

## To Reproduce

rbac enabled on kubernates
adding securityContaxt to values.yaml of the helm chart:
 securityContext:
    enabled: true
    fsGroup: 1
    runAsUser: 1001

notice that replicaset is creating a pod using the securityContext we provide,
but the spawned containers does not inherit it from the values.yaml

rajesh-kumar:
Hi, This is really crucial for us. Any updates on this, please?

davidcollom:
I believe this was resolved in #918

diemol:
Thanks @davidcollom! This will be released between tomorrow and Tuesday

PR #995, opened 2019-07-09
feat: add container security context support

### LINKED ISSUE #973: Implement container-scoped securityContext configuration
Hi, this request is similar to #904; however, in this case we'd like to have the container-scoped securityContext configurable, and passed to the spawned pods.   

This would allow us to set capabilities for the spawned containers (such as NET_ADMIN), which would in turn enable us to run tc/netem to induce impaired networking conditions in the client browser test environment.

https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.14/#securitycontext-v1-core

Please let me know if you'd like further elaboration.

Thanks,
Martin

arnaud-deprez:
Closes #973

### Description
Should fix #973 

### Motivation and Context
Allow to define [Container Security Context](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.15/#securitycontext-v1-core) for both hub and selenium grid. 
Attention the same value is reused for both Pods.

### How Has This Been Tested?
Unit tests and manual integration tests

### Types of changes

### Checklist:


arnaud-deprez:
Hi @davidcollom, @diemol, @martincfo,

Sorry for the delay, I was quite busy these days.

I did a change in #995 that is equivalent to #904 but for `Container Security Context`.
So in short, this allows to configure `Container Security Context` for both `hub` and `grid` but it does not allow to use different context/configuration.

In order to do that, I think we should foresee deeper/heavier changes. While it is still possible to implement such a feature as it is, I feel the chart and the number of environment variables to configure it all starts getting too heavy and confusing.
For example for `Pod Security Context` or `Container Security Context`, we should allow user to set json value in environment variable as it is impossible to guess all combination.

So instead of adding again some environment variable, I was thinking about using a configuration file which will look like a lot to `Pod` for the grid.

However, this idea is not clear yet in my mind as there are a lot of environment variables used for docker and docker swarm setup and so we will need to define what takes precedence over what or review the whole thing.

So basically, I propose to integrate this change as-is and then we can think about a better design/solution to fully customise zalenium deployment.

What do you think ?

diemol:
I think so far it looks good @arnaud-deprez, thanks!
If we need something more granunar in the future, then we could revise this.

PR #1016, opened 2019-07-22
Add empty default value for k8s security contexts

### LINKED ISSUE #1015: Kubernetes pods in 3.141.59t don't respect resource requests
## 🐛 Bug Report

It appears that in `3.141.59t` the environment variables that set the CPU and Memory requests/limits of the node pods in kubernetes are no longer being respected.

We have our helm values set to:
```
hub:
  # ... other hub config ...
  cpuRequest: 250m
  cpuLimit: 1000m
  memRequest: 1000Mi
  memLimit: 1500Mi
```

I can see the values being set as environment variables when describing the pod:
```
      ZALENIUM_KUBERNETES_CPU_REQUEST:     250m
      ZALENIUM_KUBERNETES_CPU_LIMIT:       1000m
      ZALENIUM_KUBERNETES_MEMORY_REQUEST:  1000Mi
      ZALENIUM_KUBERNETES_MEMORY_LIMIT:    1500Mi
```
and I confirmed they are set inside the pod using `kubectl exec`

While the nodes have the following limits (these come from the defaults for our namespace):
```
    Limits:
      cpu:     50m
      memory:  128Mi
    Requests:
      cpu:      50m
      memory:   128Mi
```

Reverting the hub to use image tag `3.141.59s` fixed our OOM problems as the limits were again set on the node.

## To Reproduce

1. Install zalenium helm chart with using the values from above
2. Use `kubectl describe` on the hub and nodes to view limits

## Expected behavior

The pods should have the expected memory/cpu limits set.

## Environment

OS: AWS EKS K8s Node
Zalenium Image Version(s): 3.141.59t

KierranM:
quickly.**

Sets empty default values for the two security contexts.

### Description

Not having these default values is causing the node pods to not have the 

### Motivation and Context

Fixes #1015 

### How Has This Been Tested?

We have tested this locally by installing the helm chart with and without the values having defaults.

### Types of changes

### Checklist:

### DIFF
--- a/charts/zalenium/values.yaml
+++ b/charts/zalenium/values.yaml
@@ -29,14 +29,14 @@ hub:
   ## Pod Security Context
   ## ref: https://kubernetes.io/docs/tasks/configure-pod-container/security-context/
   ## in case it is running outside openshift, this should fix this https://github.com/zalando/zalenium/issues/631
-  podSecurityContext:
+  podSecurityContext: {}
     ## See all the options here https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.15/#podsecuritycontext-v1-core
     # fsGroup: 1000
     # runAsUser: 1000
     # runAsGroup: 1000
 
   ## Container Security Context
-  containerSecurityContext:
+  containerSecurityContext: {}
     ## See all the options here https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.15/#securitycontext-v1-core
     # capabilities: []
     # privileged: false


KierranM:
I don't entirely understand it myself. For whatever reason when these values are left without the `{}` default the hub pod comes up with the environment variables that are required to set the requests/limits on the nodes, but the nodes don't have the requests/limits set.

As soon as you add in the default `{}` the nodes start receiving their limits again 🤷‍♀

PR #1053, opened 2019-09-18
Remove usage of transfer-logs.sh

yosserO:
### Description
The script transfer-logs.sh has been removed from browser containers but it is still used in the zalenium code. This PR removes all usages of the script transfer-logs.sh. 
https://github.com/elgalu/docker-selenium/blob/master/CHANGELOG.md#331-p9

I also made a litte fix to the function startSwarmExecContainer. The passed command needed to be parsed.

### Motivation and Context
Trying to execute a script that doesn't exist anymore in the browser container leads to the following error message in journalctl

> starting container process caused \"exec: \\\"transfer-logs.sh\\\": executable file not found in $PATH\": unknown"

Not parsing the command in the function startSwarmExecContainer leads to the following Error message in journalctl

> stat notify 'Zalenium', 'TEST COMPLETED', --icon=/home/seluser/images/completed.png: no such file or directory\": unknown"

### How Has This Been Tested?
- This has been only tested in a swarm environment
- Execute some tests
- Check that the error message listet above doesn't appear in journalctl

### Types of changes

### Checklist:

### DIFF
--- a/src/main/java/de/zalando/ep/zalenium/proxy/DockerSeleniumRemoteProxy.java
+++ b/src/main/java/de/zalando/ep/zalenium/proxy/DockerSeleniumRemoteProxy.java
@@ -604,8 +604,10 @@ public class DockerSeleniumRemoteProxy extends DefaultRemoteProxy {
     @VisibleForTesting
     void processContainerAction(final DockerSeleniumContainerAction action, final String commandParameters,
                                 final String containerId) {
-        final String[] command = { "bash", "-c", action.getContainerAction().concat(commandParameters)};
-        containerClient.executeCommand(containerId, command, action.isWaitForExecution());
+        if (DockerSeleniumContainerAction.TRANSFER_LOGS != action) {
+            final String[] command = { "bash", "-c", action.getContainerAction().concat(commandParameters)};
+            containerClient.executeCommand(containerId, command, action.isWaitForExecution());
+        }
 
         if (keepVideoAndLogs()) {
             if (DockerSeleniumContainerAction.STOP_RECORDING == action) {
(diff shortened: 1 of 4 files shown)


src/main/java/de/zalando/ep/zalenium/proxy/DockerSeleniumRemoteProxy.java
```diff
@@ -611,9 +611,6 @@ void processContainerAction(final DockerSeleniumContainerAction action, final St
             if (DockerSeleniumContainerAction.STOP_RECORDING == action) {
                 copyVideos(containerId);
             }
-            if (DockerSeleniumContainerAction.TRANSFER_LOGS == action) {
```
diemol: @yosserO I know that the `transfer-logs.sh` does not exist in the `elgalu/selenium` image, and it is not correct to invoke the action a few lines [above](https://github.com/zalando/zalenium/pull/1053/files#diff-cdb17f721a30ede405aeccaa422c3006L608), but removing this `if` basically means that no logs will be copied when someone is running Zalenium in k8s or docker in a single host. So deleting this would break it for those environments.

Perhaps a fix to avoid executing that command unnecessarily would be to avoid executing lines [607-608](https://github.com/zalando/zalenium/pull/1053/files#diff-cdb17f721a30ede405aeccaa422c3006L607) when the action is `TRANSFER_LOGS`?

yosserO: @diemol , thank you for you feedback. My Bad I didn't consider the other environments when doing the changes. I will revert them and wrap the lines 607-608 with an if condition to avoid executing the script when the action is TRANSFER_LOGS.

PR #1059, opened 2019-09-30
Initial implemenation of custom nginx error pages

TobiX:
### Description

This adds a simple Zalenium-specific error page (idea from https://blog.adriaan.io/one-nginx-error-page-to-rule-them-all.html), nothing fancy, maybe someone else wants to add some style.

grafik

### Motivation and Context

When debugging issues in a Kubernetes cluster, you always see the default nginx error pages, because nginx is everywhere. Sometime it's hard to know which system created the error, so this makes it possible at first glance to see if the error was in Zalenium or Kubernetes.

### How Has This Been Tested?

Built a container, but only manually tested that the nginx config works.

### Types of changes

### Checklist:

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -366,7 +366,7 @@ ENV ZAL_VER="${project.build.finalName}" \
     SELENIUM_WAIT_FOR_CONTAINER="true"
 
 COPY entry.sh log warn error /usr/bin/
-COPY nginx.conf /home/seluser/
+COPY nginx.conf error.html /home/seluser/
 COPY css/ /home/seluser/css/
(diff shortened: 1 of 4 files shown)


TobiX:
Added a screenshot to the pull request description. As I said, only a little bit more then the nginx error pages, maybe someone else wants to add some style.

I pushed a very minor fix, since my error pages provoked a subtle bug:
When checking if the Hub has started correcty, JQ now thre the error
```
parse error: Invalid numeric literal at line 1, column 10
```
instead of
```
parse error: Invalid numeric literal at line 1, column 7
```
which matched the following grep...

PR #1073, opened 2019-10-26
EKS scaleup issues

diego-fernandez-santos:
When you try to scale up Zalenium to more than 400 pods some concurrency and race conditions issues appears.

* Avoid race conditions in registration process
	- Kubernetes has eventual consistency, Zalenium can receive an registration request from a POD which is still not ready in eks
	- TearDown already registered nodes at ZaleniumRegistry

	ZaleniumRegistry.add
	AutoStartProxySet.register & checkContainers

* Increase HTTP Timeouts to avoid problems under load
	seconds -> milliseconds

*  Avoid java.lang.OutOfMemoryError: unable to create new native thread
	Use thread pool in AutoStartProxySet start method.

*  Separate stale and idle timeout. Differentiate two times of timeouts, one for sessions without activity a specific interval of time (stale/idle timeout)  and other for nodes without activity (pool timeout)
	DockerSeleniumRemoteProxy

*  Avoid get stuck while reading stream - Add waitForInputStreamToConnect timeout
	KubernetesContainerClient

* Several checkContainers and register race conditions
	Pod is register while shutdowning node registration took too long
	Pod is not mark has idle is connection is closed by the client while it was waiting for a session.
	Pod is marked as stuck even when the number of pods is equal or lower to the minimum.
	Differentiate two times of timeouts, one for sessions without activity a specific interval of time (stale/idle timeout)  and other for nodes without activity (pool timeout) to avoid kill pods which are still scaling (in eks the scaleup can take more than 90s).

* Avoid file names too long and repeat process same requests
	Prefix custom capabilities and truncate them when they are longer than 250 characters in addNewSessionRequest

### DIFF
--- a/src/main/java/de/zalando/ep/zalenium/container/kubernetes/KubernetesContainerClient.java
+++ b/src/main/java/de/zalando/ep/zalenium/container/kubernetes/KubernetesContainerClient.java
@@ -4,16 +4,20 @@ import de.zalando.ep.zalenium.container.ContainerClient;
 import de.zalando.ep.zalenium.container.ContainerClientRegistration;
 import de.zalando.ep.zalenium.container.ContainerCreationStatus;
 import de.zalando.ep.zalenium.util.Environment;
+import io.fabric8.kubernetes.api.model.BaseKubernetesListFluent.KubernetesClusterRoleItemsNested;
+import io.fabric8.kubernetes.api.model.ContainerStateRunning;
 import io.fabric8.kubernetes.api.model.Container;
 import io.fabric8.kubernetes.api.model.ContainerStateTerminated;
 import io.fabric8.kubernetes.api.model.ContainerStatus;
 import io.fabric8.kubernetes.api.model.DoneablePod;
 import io.fabric8.kubernetes.api.model.EnvVar;
 import io.fabric8.kubernetes.api.model.HostAlias;
+import io.fabric8.kubernetes.api.model.IntOrString;
 import io.fabric8.kubernetes.api.model.LocalObjectReference;
 import io.fabric8.kubernetes.api.model.Pod;
 import io.fabric8.kubernetes.api.model.PodList;
 import io.fabric8.kubernetes.api.model.PodSecurityContext;
+import io.fabric8.kubernetes.api.model.PodStatus;
 import io.fabric8.kubernetes.api.model.Quantity;
 import io.fabric8.kubernetes.api.model.SecurityContext;
 import io.fabric8.kubernetes.api.model.Toleration;
@@ -40,6 +44,7 @@ import java.util.Map;
 import java.util.Optional;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.function.Function;
 import java.util.function.Supplier;
@@ -414,6 +419,26 @@ public class KubernetesContainerClient implements ContainerClient {
         }
     }
 
+    public PodStatus getPodStatus(ContainerCreationStatus container) {
+        Pod pod = client.pods().withName(container.getContainerName()).get();
+        if (pod == null) {
+            logger.debug("Pod not found {}", container);
+            return null;
+        } else {
+            return pod.getStatus();
+        }
+    }   
+
+    public String getContainerIdByIp(String PodIP) {
+        for(Pod pod : client.pods().list().getItems()) {
+            if (PodIP != null && PodIP.equals(pod.getStatus().getPodIP())) {
+                return pod.getMetadata().getName();
+            }
+        }
+        logger.debug("Pod not found by Ip {}", PodIP);
+        return null;
+    }   
+
     public boolean isTerminated(ContainerCreationStatus container) {
         Pod pod = client.pods().withName(container.getContainerName()).get();
         if (pod == null) {
@@ -450,9 +475,13 @@ public class KubernetesContainerClient implements ContainerClient {
         for (Pod pod : list.getItems()) {
 
             if (podIpAddress.equals(pod.getStatus().getPodIP())) {
+                if ("Running".equals(pod.getStatus().getPhase())) {
                 containerId = pod.getMetadata().getName();
                 currentPod = pod;
                 break;
+                } else {
+                    logger.warn("Pod {} ip duplicated", pod.getMetadata().getName());
+                }
             }
         }
 
@@ -529,7 +558,7 @@ public class KubernetesContainerClient implements ContainerClient {
 
         public void waitForInputStreamToConnect() {
             try {
-                this.openLatch.await();
+                this.openLatch.await(10, TimeUnit.SECONDS);
             }
             catch (InterruptedException e) {
                 logger.error( String.format("%s Failed to execute command %s", containerId, Arrays.toString(command)), e);
@@ -610,16 +639,24 @@ public class KubernetesContainerClient implements ContainerClient {
                         .endResources()
                         // Add a readiness health check so that we can know when the selenium pod is ready to accept requests
                         // so then we can initiate a registration.
-                        .withNewReadinessProbe()
-                            .withNewExec()
-                                .addToCommand(new String[] {"/bin/sh", "-c", "http_proxy=\"\" curl -s http://`hostname -i`:"
-                                        + config.getNodePort() + "/wd/hub/status | jq .value.ready | grep true"})
-                            .endExec()
-                            .withInitialDelaySeconds(5)
-                            .withFailureThreshold(60)
-                            .withPeriodSeconds(1)
+                        .withNewLivenessProbe()
+                            .withInitialDelaySeconds(30)
                             .withTimeoutSeconds(5)
+                            .withNewHttpGet()
+                                .withPath("/wd/hub/status")
+                                .withPort(new IntOrString(40000))
+                            .endHttpGet()
+                        .endLivenessProbe()
+                        .withNewReadinessProbe()
+                            .withInitialDelaySeconds(5)
+                            .withFailureThreshold(2)
+                            .withPeriodSeconds(2)
+                            .withTimeoutSeconds(10)
                             .withSuccessThreshold(1)
+                            .withNewHttpGet()
+                                .withPath("/wd/hub/status")
+                                .withPort(new IntOrString(40000))
+                            .endHttpGet()
                         .endReadinessProbe()
                     .endContainer()
                     .withRestartPolicy("Never")
--- a/src/main/java/de/zalando/ep/zalenium/container/kubernetes/KubernetesContainerClient.java
+++ b/src/main/java/de/zalando/ep/zalenium/container/kubernetes/KubernetesContainerClient.java
@@ -469,11 +469,13 @@ public class KubernetesContainerClient implements ContainerClient {
         String podIpAddress = remoteHost.getHost();
 
         // The only way to lookup a pod name by IP address is by looking at all pods in the namespace it seems.
-        PodList list = client.pods().withLabels(createdByZaleniumMap).list();
+        // We reduced the resiterCycle to 1-10 seconds and change it to use a cache
+        // to avoid overload kubernetes
+        refreshCachedPodList();
 
         String containerId = null;
         Pod currentPod = null;
-        for (Pod pod : list.getItems()) {
+        for (Pod pod : cachedPodList.values()) {
 
             if (podIpAddress.equals(pod.getStatus().getPodIP())) {
                 String podPhase = pod.getStatus().getPhase();
@@ -511,7 +513,7 @@ public class KubernetesContainerClient implements ContainerClient {
         return registration;
     }
 
-    private Pod getCachedPodWithName(String name) {
+    private void refreshCachedPodList() {
         long now = System.currentTimeMillis();
         synchronized(this) {
             if (now - cachedPodListLastUpdated > cachePodListTime) {
@@ -526,6 +528,11 @@ public class KubernetesContainerClient implements ContainerClient {
                         cachedPodList.size());
             }
         }
+    }
+
+    private Pod getCachedPodWithName(String name) {
+        long now = System.currentTimeMillis();
+        refreshCachedPodList();
         Pod pod = cachedPodList.get(name);
         if (pod == null) {
             logger.warn("Pod {} was not in the cache", name);
--- a/src/main/java/de/zalando/ep/zalenium/container/kubernetes/KubernetesContainerClient.java
+++ b/src/main/java/de/zalando/ep/zalenium/container/kubernetes/KubernetesContainerClient.java
@@ -40,6 +40,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicBoolean;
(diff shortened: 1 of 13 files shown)


src/main/java/de/zalando/ep/zalenium/container/kubernetes/KubernetesContainerClient.java
```diff
@@ -414,6 +419,26 @@ public boolean isReady(ContainerCreationStatus container) {
         }
     }
 
+    public PodStatus getPodStatus(ContainerCreationStatus container) {
```
diemol: What is the purpose of this `getPodStatus` method? I see it being used in the `DockeredSeleniumStarter.java` class, but the method `getPodStatus` over there is not called anywhere else.

src/main/java/de/zalando/ep/zalenium/container/kubernetes/KubernetesContainerClient.java
```diff
@@ -450,9 +475,13 @@ public ContainerClientRegistration registerNode(String zaleniumContainerName, UR
         for (Pod pod : list.getItems()) {
 
             if (podIpAddress.equals(pod.getStatus().getPodIP())) {
+                if ("Running".equals(pod.getStatus().getPhase())) {
```
diemol: Would it be better to do a lower case comparison? In case one day the API response changes?

diemol: Also, it'd be nice to keep the indentation.

src/main/java/de/zalando/ep/zalenium/container/kubernetes/KubernetesContainerClient.java
```diff
@@ -529,7 +558,7 @@ public void onClose(int code,
 
         public void waitForInputStreamToConnect() {
             try {
-                this.openLatch.await();
+                this.openLatch.await(10, TimeUnit.SECONDS);
```
diemol: What is the reasoning behind this change?

src/main/java/de/zalando/ep/zalenium/container/kubernetes/KubernetesContainerClient.java
```diff
@@ -610,16 +639,24 @@ public static DoneablePod createDoneablePodDefaultImpl(PodConfiguration config)
                         .endResources()
                         // Add a readiness health check so that we can know when the selenium pod is ready to accept requests
                         // so then we can initiate a registration.
+                        .withNewLivenessProbe()
```
diemol: Why do we need a liveness probe?

src/main/java/de/zalando/ep/zalenium/container/kubernetes/KubernetesContainerClient.java
```diff
@@ -610,16 +639,24 @@ public static DoneablePod createDoneablePodDefaultImpl(PodConfiguration config)
...
                         .withNewReadinessProbe()
-                            .withNewExec()
-                                .addToCommand(new String[] {"/bin/sh", "-c", "http_proxy=\"\" curl -s http://`hostname -i`:"
-                                        + config.getNodePort() + "/wd/hub/status | jq .value.ready | grep true"})
-                            .endExec()
                             .withInitialDelaySeconds(5)
-                            .withFailureThreshold(60)
-                            .withPeriodSeconds(1)
-                            .withTimeoutSeconds(5)
+                            .withFailureThreshold(2)
```
diemol: Why are there changes in the values? I'd be nice to understand the reasoning behind that.

src/main/java/de/zalando/ep/zalenium/container/kubernetes/KubernetesContainerClient.java
```diff
@@ -610,16 +639,24 @@ public static DoneablePod createDoneablePodDefaultImpl(PodConfiguration config)
...
+                            .withPeriodSeconds(2)
+                            .withTimeoutSeconds(10)
                             .withSuccessThreshold(1)
+                            .withNewHttpGet()
```
diemol: We would still need to know if the JSON payload from the status endpoint returns ready or not, right?

src/main/java/de/zalando/ep/zalenium/proxy/DockerSeleniumRemoteProxy.java
```diff
@@ -66,7 +67,7 @@
     @VisibleForTesting
     public static final String ZALENIUM_MAX_TEST_SESSIONS = "ZALENIUM_MAX_TEST_SESSIONS";
     @VisibleForTesting
-    public static final long DEFAULT_MAX_TEST_IDLE_TIME_SECS = 90L;
+    public static final long DEFAULT_MAX_TEST_IDLE_TIME_SECS = 300L;
```
diemol: This value can be modified through an environment variable, right? Why do we need to change it here?

src/main/java/de/zalando/ep/zalenium/proxy/DockerSeleniumRemoteProxy.java
```diff
@@ -451,30 +452,54 @@ public synchronized boolean isTestSessionLimitReached() {
...
     public boolean shutdownIfIdle() {
         String currentName = configureThreadName();
-        boolean testIdle = isTestIdle();
-        boolean testSessionLimitReached = isTestSessionLimitReached();
-        boolean isShutdownIfIdle = testIdle || (testSessionLimitReached && !isBusy());
-        if (isShutdownIfIdle) {
-            LOGGER.debug("Proxy is idle.");
-            timeout("proxy being idle after test.", (testSessionLimitReached ?
-                    ShutdownType.MAX_TEST_SESSIONS_REACHED : ShutdownType.IDLE));
+        boolean condition = isTestIdleMoreThan(900) && !isCleaningUp() && !isBusy();
```
diemol: Could you please elaborate on the logic here? Why 900 seconds? Are we not checking if the test session limit is reached anymore?

src/main/java/de/zalando/ep/zalenium/proxy/DockerSeleniumRemoteProxy.java
```diff
@@ -451,30 +452,54 @@ public synchronized boolean isTestSessionLimitReached() {
...
         String currentName = configureThreadName();
-        if (isBusy() && isTestIdle() && !isCleaningUp()) {
-            LOGGER.debug("No test activity been recorded recently, proxy is stale.");
-            timeout("proxy being stuck | stale during a test.", ShutdownType.STALE);
+        boolean condition = isTestIdle() && !isCleaningUp() && isBusy();
+        if (condition) {
+            String msg = String.format("Test did not see a new command "
+                    + "for %s seconds. Timing out", getMaxTestIdleTimeSecs());
+            LOGGER.debug(msg);
+            timeout(msg, ShutdownType.STALE);
```
diemol: Why don't we report the `MAX_TEST_SESSIONS_REACHED` shutdown type anymore? Could you please elaborate on the logical change here?

src/main/java/de/zalando/ep/zalenium/proxy/DockerSeleniumRemoteProxy.java
```diff
@@ -451,30 +452,54 @@ public synchronized boolean isTestSessionLimitReached() {
...
+        }
             setThreadName(currentName);
-            return true;
-        } else if (isTestSessionLimitReached() && !isBusy()) {
-            LOGGER.debug("Proxy has reached max test sessions.");
-            timeout("proxy has reached max test sessions.", ShutdownType.MAX_TEST_SESSIONS_REACHED);
+        return condition;
+    }    
+
+    public boolean shutdownIfCorrupt(String containerId) {
```
diemol: What does it mean for a container to be corrupt? Would it be possible that it is still starting, it has an IP assigned but it is not ready yet?

src/main/java/de/zalando/ep/zalenium/registry/ZaleniumRegistry.java
```diff
@@ -251,8 +264,24 @@ public void addNewSessionRequest(RequestHandler handler) {
         try {
             lock.lock();
             Map<String, Object> requestedCapabilities = handler.getRequest().getDesiredCapabilities();
-            proxies.verifyAbilityToHandleDesiredCapabilities(requestedCapabilities);
+
+            //Prefix custom capabilities and truncate them when they are longer than 250 characters
+            for(String capability :  Arrays.asList("name" , "zal:name", "build", "zal:build")) {
```
diemol: The prefixing should be done at the client, for example C# does not allow you to send a test request if custom capabilities are not prefixed.

src/main/java/de/zalando/ep/zalenium/proxy/AutoStartProxySet.java
```diff
@@ -600,3 +645,34 @@ public String toString() {
...
+
+    @Override
+    public V remove(Object key) {
+        StringBuilder stackTrace = new StringBuilder();
+        Arrays.asList(Thread.currentThread().getStackTrace())
+            .forEach( (val) -> stackTrace.append(val));
+
+        LOGGER.info(String.format("TRACK: Remove proxy %s\n%s",
+                key.toString(),
+                stackTrace.toString()));
```
pearj: This looks like debugging to me, I think it should use LOGGER.debug, and then wrap in an if statement checking if debug is enabled, because getting stack traces is usually pretty expensive.

src/main/java/de/zalando/ep/zalenium/container/kubernetes/KubernetesContainerClient.java
diego-fernandez-santos: Yeah, I think that now I can remove it, it's only for debug. I'll remove it.

src/main/java/de/zalando/ep/zalenium/container/kubernetes/KubernetesContainerClient.java
diego-fernandez-santos: Ok

src/main/java/de/zalando/ep/zalenium/container/kubernetes/KubernetesContainerClient.java
diego-fernandez-santos: In some cases the thread got waiting for ever.

src/main/java/de/zalando/ep/zalenium/registry/ZaleniumRegistry.java
diego-fernandez-santos: We didn't prefix nothing in the client but we found that Zalenium  somewhere added the prefix zal: to some of the capabilities and then it started to process the same request several times. We try to find the specific case.

src/main/java/de/zalando/ep/zalenium/proxy/DockerSeleniumRemoteProxy.java
diego-fernandez-santos: Yes, we should remove it from here.

src/main/java/de/zalando/ep/zalenium/container/kubernetes/KubernetesContainerClient.java
diego-fernandez-santos: Yes, but we have less problem using this way to check the status although there are cases where the grid can not be ready. . With high concurrency the readiness didn't work properly.

src/main/java/de/zalando/ep/zalenium/proxy/AutoStartProxySet.java
diego-fernandez-santos: Yes, if was for debugging, we're going to remove it

src/main/java/de/zalando/ep/zalenium/container/kubernetes/KubernetesContainerClient.java
diego-fernandez-santos: You're right, we don't need the livenessprobe, we'll also remove it.

src/main/java/de/zalando/ep/zalenium/proxy/DockerSeleniumRemoteProxy.java
diego-fernandez-santos: It's reported in the shutdownIfLimitReached

src/main/java/de/zalando/ep/zalenium/proxy/DockerSeleniumRemoteProxy.java
diego-fernandez-santos: This specific timeout of 900 seconds it was to differentiate also another scenario.  For me it's not the same the case where an active session stop to receive commands to the case where the session is closed. We have a penalty of pod creation of about 180 seconds, so

src/main/java/de/zalando/ep/zalenium/proxy/DockerSeleniumRemoteProxy.java
diego-fernandez-santos: Sorry, I just realize that that timeout can't  be modified using environment variables, you only can change it using the request capabilities.

diemol: Yes, you are right. I confused it with other value. 
But then, the better approach would be to make this value configurable, right?

src/main/java/de/zalando/ep/zalenium/container/kubernetes/KubernetesContainerClient.java
diego-fernandez-santos: About the unit tests, I realized that the problem is because in the test environment the pods doesn't have pod phase. So now, with the new running phase check, the registerNode fail.

We added this running phase check because a selenium-nodes (pods) can try to register   before the eks API Server knows he is in Running Phase what later will create other problems, i.e. in checkContainers. If the registration process fail the selenium-node automatically is going to try to register later so this check solve that problem.

Do you know more about how to avoid this problem in the Test environment? If you want, we can try to talk about this by phone.

src/main/java/de/zalando/ep/zalenium/proxy/DockerSeleniumRemoteProxy.java
diego-fernandez-santos: done

src/main/java/de/zalando/ep/zalenium/container/kubernetes/KubernetesContainerClient.java
diemol: Should we leave this configurable then?

src/main/java/de/zalando/ep/zalenium/container/kubernetes/KubernetesContainerClient.java
diemol: Hi, could you please elaborate on the previous question?

src/main/java/de/zalando/ep/zalenium/container/kubernetes/KubernetesContainerClient.java
diemol: But the only way to know if the node is ready is to query the endpoint, if the enpoint is not ready then the node cannot receive any tests. Am I missing something?

src/main/java/de/zalando/ep/zalenium/proxy/DockerSeleniumRemoteProxy.java
diemol: Yeah, but how can we keep the previous behaviour as well?

src/main/java/de/zalando/ep/zalenium/proxy/DockerSeleniumRemoteProxy.java
diemol: Shouldn't that be fixed in the hub? By removing pods that die and registering the new one?

src/main/java/de/zalando/ep/zalenium/registry/ZaleniumRegistry.java
diemol: Yes, we add the prefix as a way to help some bindings, but in the end, this should be done in the client.

diemol:
Hi, I will close this PR since it went stale and also because the project is not under development anymore.

PR #1080, opened 2019-11-05
feature(): use custom scheduller in k8s

alfonso-presa:
### Description
This change allows using a custom scheduler in EKS both for the hub and for the bootstrapped browsers.

### Motivation and Context
In some cloud platforms like EKS o GKE, the scheduling algorithm performed by the default scheduler of k8s cannot be tuned. We have found that in order to have reasonable dynamic up scaling and down scaling of the cluster when you have warmed pods, and high varying browser concurrency (with changes from 20 browsers to up to 400), you need to use [`MostRequestedPriority`](https://kubernetes.io/docs/concepts/scheduling/) in order for nodes to get deleted by cluster autoscaler. This change enables setting a custom scheduler name for pods owned by zalenium so that this is possible.

### How Has This Been Tested?
We have made tests with a grid with 30 warmed up browsers in EKS, which initially required 4 nodes. **Without the changes**, we gave the grid big demand of 200 browsers, which brought up 20 nodes. After tests ended the grid was down scaled to 15 nodes, and the extra 11 nodes were never released (as they where occupied by warmed up browsers).
**With the changes**, we gave the grid big demand of 200 browsers, which brought up 20 nodes. After tests ended the grid was down scaled to 6 nodes. (I will PR latter with another solution to make them shrink down to the initial 4 nodes).

### Types of changes

### Checklist:
- [] I have updated the documentation accordingly.
- [X] I have added tests to cover my changes.

### DIFF
--- a/charts/zalenium/templates/_pod-template.yaml
+++ b/charts/zalenium/templates/_pod-template.yaml
@@ -5,6 +5,10 @@ metadata:
   annotations:
     {{- toYaml .Values.hub.podAnnotations | nindent 4 }}
 spec:
+  {{- if (.Values.hub.schedulerName) }}
+  schedulerName: {{ .Values.hub.schedulerName }}
+  {{- end}}
+
   securityContext:
     {{- toYaml .Values.hub.podSecurityContext | nindent 4 }}
   {{- if .Values.hub.pullSecrets }}
--- a/charts/zalenium/values.yaml
+++ b/charts/zalenium/values.yaml
@@ -75,6 +75,12 @@ hub:
   serviceAnnotations: {}
 
   # This enables using a custom scheduler for both hub and browser pods
+  # This could be useful to have a cost effective pod distribution with
+  # warmmed up browsers in k8s nodes by using a custom scheduler
+  # (i.e. by deploying a new one in your cluster) configured with
+  # MostRequestedPriority (https://kubernetes.io/docs/concepts/scheduling/)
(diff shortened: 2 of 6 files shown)


alfonso-presa:
Hi @diemol 

thank you. Let me know if you need any more info for those tests.

About documentation, do you mean including something more specific to the purpose of the value `schedulerName`? I added some little info, but you're right that it might be useful to give some pointers to readers on how to use it for better pod distribution and infrastructure cost optimization. Should I add this to the chart? I din't found any other place to add k8s related documentation, but if you point it out to me, I would be pleased to contribute there.

Thanks!

diemol:
I think in the helm chart should be fine

alfonso-presa:
Hi @diemol, I've added some tips to the values.yaml. Please check if looks good and enough.

diemol:
Sorry, I broke the commit history of the PR. But no worries, I will merge this as it is and then run all CI in a protected branch.
Thanks @alfonso-presa!

alfonso-presa:
No problem. If you prefer I could close this one and reopen 😉.

diemol:
I am not sure what happened but now the history looks fine 🤷‍♂

alfonso-presa:
It might be that target branch was initially changed to an outdated one missing some commit from master that were coming along with my branch based in master? If you updated the target branch later on with master commits those commits could have initially be included in my PR and then removed when they already arrived to master via the update.

Issue #490, opened 2018-03-20
Ability to start and stop video recording from the tests.


sri85:
We use protractor for our e2e tests , currently we are using vanilla selenium grid in our pipelines to run our tests but we are planning to switch to Zalenium (because its awesome) . We do run our tests in parallel and the way protractor runs the tests in parallel is that it shards the feature file and each feature file is run in a different browser. But the issue that we are having is when we record a video Zalenium creates one video file per feature file , but then this artifact might not be very useful when the developer views the video to debug the issue(Lets say if the feature file has 4 scenarios , if the third tests fails , the developer will need to see videos of first 3 scenarios which have passed)

Is there a way I could programmatically control the video recording via an API , if its available then possibly i could use a `cucumber` `AfterSceario` `BeforeScenario` hook to start and stop the video after every scenario

iMbaundso:
Please have a look at https://github.com/zalando/zalenium/issues/471.

We're also using Protractor (jasmine instead of cucumber) with multicaps for our e2e tests and do not acquire a new Zalenium session between every scenario. We do however have an additional reporting which says which scenario has failed including the exception. This made it quite easy to fast forward the video to the interesting parts once you've done it a few times. 
Additionally we are doing a some kind of visually marking the first error using protractors browser.executeScript in a jasmine callback (we are making our page header red which makes it quite obvious to see at what point the test failed ;)). You could do something similar using cucumber hooks I guess (maybe have a look at some reporters as they are doing something similar)

Protractor snippet could look something like this:
browser.executeScript('document.querySelector(".header").style.background = red');

diemol:
Hi @sri85 @iMbaundso @gpapadakis

Actually, I was thinking that we could do something about it for those types of tests. Maybe an alternative would be to keep using cookies to send custom commands to Zalenium (as when a test passes or a message is displayed on screen). Perhaps a cookie saying `new Cookie("zaleniumVideo", "START");` and `new Cookie("zaleniumVideo", "STOP");`.

It is just an idea, and I am kind of short of time these days since the free time I have, we are trying to improve the Kubernetes part. But if someone feels like implementing it, we would be very happy to accept PRs.

sri85:
Thanks for the idea @diemol  ,  i will try to create a PR

javacoldbrew:
Has there been any update on this?

diemol:
Closing, as it is consolidated in #645
If development starts, we'll reopen the issue.

gpapadakis:
[...]

PR #1085, opened 2019-11-09
Add feature to start or stop video recording using cookie

ccharnkij:
### Description
This changes allow users to start or stop video recording by setting the  cookie named 'zaleniumVideo' to either 'true' or 'false'. This can be done at any time regardless of what the environment variable or the session variable is set to.

### Motivation and Context
Per #490 , this is to assist with troubleshooting when multiple tests are run in the same session. Typically, when multiple tests are using the same session, all tests are recorded in a single video file. This makes it inconvenient to locate the failed tests. With this change, users can split each test into its own video file, even when they are in the same session.

Also to further assist with identifying the test, a cookie named 'zaleniumTestName' can be passed in to set the test name. This is so that each test that gets split up can have its own unique name. If name already exists, a count will be appended to the name.

### How Has This Been Tested?
Unit tests have been added. Also functional UI test using WebDriverIO-Cucumber were used. I have tried starting the hub with video recording enabled or disable, and then I triggered the scenario outline test with multiple examples while splitting up each recording into its own file. 

### Types of changes

### Checklist:

### Extra Note:
This feature can be used for splitting up ONLY video file. The log files only appear with the last video. This is because I cannot alter the logging behavior of the Selenium node container from @elgalu from within Zalenium.

### DIFF
--- a/src/main/java/de/zalando/ep/zalenium/proxy/DockerSeleniumRemoteProxy.java
+++ b/src/main/java/de/zalando/ep/zalenium/proxy/DockerSeleniumRemoteProxy.java
@@ -104,6 +104,7 @@ public class DockerSeleniumRemoteProxy extends DefaultRemoteProxy {
     private long cleanupStartedTime = 0;
     private AtomicBoolean timedOut = new AtomicBoolean(false);
     private long timeRegistered = System.currentTimeMillis();
+    private boolean stopRecordingByCookie = false;
 
     public DockerSeleniumRemoteProxy(RegistrationRequest request, GridRegistry registry) {
         super(request, registry);
@@ -323,6 +324,27 @@ public class DockerSeleniumRemoteProxy extends DefaultRemoteProxy {
                         processContainerAction(DockerSeleniumContainerAction.CLEAN_NOTIFICATION, getContainerId());
                         processContainerAction(DockerSeleniumContainerAction.SEND_NOTIFICATION, messageCommand,
                                 getContainerId());
+                    }
+                    if ("zaleniumVideo".equalsIgnoreCase(cookieName)) {
+                        boolean recordVideo = Boolean.parseBoolean(cookie.get("value").getAsString());
+                        if (recordVideo && !isVideoRecordingEnabled()) {
+                            setVideoRecordingEnabledSession(true);
+                            testInformation.setVideoRecorded(true);
+                            stopRecordingByCookie = false;
+                            videoRecording(DockerSeleniumContainerAction.START_RECORDING);
+                        } else if (!recordVideo && isVideoRecordingEnabled()){
+                            stopRecordingByCookie = true;
+                            videoRecording(DockerSeleniumContainerAction.STOP_RECORDING);
+                            setVideoRecordingEnabledSession(false);
+                            testInformation.setVideoRecorded(false);
+                        }
+                    }
+                    if ("zaleniumTestName".equalsIgnoreCase(cookieName)) {
+                        String newTestName = cookie.get("value").getAsString();
+                        if (!newTestName.isEmpty()) {
+                            testName = newTestName;
+                            testInformation.setTestName(testName);
+                        }
                     }
                     else if(CommonProxyUtilities.metadataCookieName.equalsIgnoreCase(cookieName)) {
                         JsonParser jsonParser = new JsonParser();
@@ -632,6 +654,9 @@ public class DockerSeleniumRemoteProxy extends DefaultRemoteProxy {
         if (keepVideoAndLogs()) {
             if (DockerSeleniumContainerAction.STOP_RECORDING == action) {
                 copyVideos(containerId);
+                if (stopRecordingByCookie) {
+                    DashboardCollection.updateDashboard(testInformation);
+                }
             }
             if (DockerSeleniumContainerAction.TRANSFER_LOGS == action) {
                 copyLogs(containerId);
@@ -658,9 +683,18 @@ public class DockerSeleniumRemoteProxy extends DefaultRemoteProxy {
                 if (!Files.exists(videoFile.getParent())) {
                     Files.createDirectories(videoFile.getParent());
                 }
+
+                //For multiple recording in one session, add '_{count}' to the test name
+                if (Files.exists(videoFile)) {
+                    testInformation.setTestName(testName + "_" + testInformation.getFileCount());
+                    testInformation.buildVideoFileName();
+                    videoFile = Paths.get(String.format("%s/%s", testInformation.getVideoFolderPath(),
+                            testInformation.getFileName()));
+                }
                 Files.copy(entry.get(), videoFile);
                 CommonProxyUtilities.setFilePermissions(videoFile);
                 videoWasCopied = true;
+                testInformation.setFileCount(testInformation.getFileCount() + 1);
                 LOGGER.debug("Video file copied to: {}/{}", testInformation.getVideoFolderPath(), testInformation.getFileName());
             }
         } catch (IOException e) {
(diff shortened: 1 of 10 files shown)


src/main/java/de/zalando/ep/zalenium/proxy/DockerSeleniumRemoteProxy.java
```diff
@@ -769,8 +805,8 @@ private void cleanupNode(boolean willShutdown) {
     }
 
     private boolean keepVideoAndLogs() {
-        return !keepOnlyFailedTests || TestInformation.TestStatus.FAILED.equals(testInformation.getTestStatus())
-                || TestInformation.TestStatus.TIMEOUT.equals(testInformation.getTestStatus());
+        return isVideoRecordingEnabled() && (!keepOnlyFailedTests || TestInformation.TestStatus.FAILED.equals(testInformation.getTestStatus())
```
diemol: Why do we need to add this condition now?

src/test/java/de/zalando/ep/zalenium/proxy/DockerSeleniumRemoteProxyTest.java
```diff
@@ -557,6 +557,76 @@ public void videoRecordingIsDisabledViaCapability() {
       Assert.assertFalse(proxy.isVideoRecordingEnabled());
     }
 
+    @Test
```
diemol: Can we add an end to end test as well please?

ccharnkij:
Hello @diemol,

Thank you for getting back to me.

- How can I mark test status for each video recording?
This can be done via the cookie 'zaleniumTestPassed'. A user can set this cookie before 'zaleniumVideo'.
- How can I be sure that test names are well differentiated in the files?
A user can set the cookie 'zaleniumTestName'. If not, I append the default file name with '_{count}'.
default_file_name
set_file_name_and_test_status
- The dashboard is showing each video as its own test, is this intended?
I'm not sure what this means. This feature is only meant to help with debugging in the case that a user uses a framework/library that run multiple tests in one session.
- What is the overall use case? I am missing that.
This is to help with something like #490. Libraries such as Cucumber w/ WebDriverIO run every test inside a feature file in one session. So a user would get one video with multiple tests in it. A user would have to forward/rewind the video to a specific point to see a specific test. With this cookie, a user can sectioning out only the interested portion.
- Can we add documentation please?
Sure. Can I do that last, once (and if) this PR is good? everything is in the docs directory, yes?
- How would this work with the cloud integrations? We cannot support this feature if someone runs a test in Sauce Labs for example...
Good Point, this is not in scope for this feature. I only planned this out for elgalu Selenium.

src/main/java/de/zalando/ep/zalenium/proxy/DockerSeleniumRemoteProxy.java
ccharnkij: Looking at this again, this doesn't seem necessary. I can change it back. I think I added to left side so that it doesn't have to evaluate the right side if it doesn't have to.

src/test/java/de/zalando/ep/zalenium/proxy/DockerSeleniumRemoteProxyTest.java
ccharnkij: Sure, can you point me to some example?

src/main/java/de/zalando/ep/zalenium/proxy/DockerSeleniumRemoteProxy.java
diemol: OK, then let's leave it the way it was 🤝

src/test/java/de/zalando/ep/zalenium/proxy/DockerSeleniumRemoteProxyTest.java
diemol: End to End tests are here: https://github.com/zalando/zalenium/blob/master/src/test/java/de/zalando/ep/zalenium/it/ParallelIT.java

Just add one that only runs on Zalenium and uses the added cookie names. Afterwards it should check that the amount of videos matches the amount of tests.

diemol:
@ccharnkij yes, the PR looks good, but then we need to add documentation explaining the use case and its scope.

ccharnkij:
@diemol would you like me to rebase all the recent commits that I didn't have and push again?

diemol:
@ccharnkij have you seen that there are merge conflicts? Please have a look and then we can merge this PR. Thanks!

diemol:
Seems something went wrong with the merge conflicts fix @ccharnkij, can you please have a look?
I see many unrelated commits.

diemol:
@ccharnkij I executed all tests in PR #1133 and the integration test is failing
https://travis-ci.org/zalando/zalenium/jobs/638147676#L3969
Could you please have a look?

ccharnkij:
@diemol It looks like the tests were run via saucelab, is this correct? If so, I would expect it to fail since this feature only works with elgalu/selenium. I've run it again with elgalu/selenium and it was fine.

robeves95:
This is a great idea and will be very useful. Any idea when it will be available as part of the latest Zalenium image please?

diemol:
@ccharnkij not really, it is only using `elgalu/selenium`. I think the problem is that the [assertion](https://github.com/zalando/zalenium/blob/1508af9cdd4b6b69716d90e73960cd9dae0bb467/src/test/java/de/zalando/ep/zalenium/it/ParallelIT.java#L224) is not waiting for the dashboard to have the elements present. Perhaps you should add an expected condition?

ccharnkij:
@diemol I've added wait condition to the IT. Please try again.

ccharnkij:
@diemol Yes, I am aware of the duplicate file. The reason is because of the method cleanupNode in DockerSeleniumRemoteProxy, specifically the updateDashboard line. This method is triggered at the end by afterSession and put results in the dashboard no matter what, regardless of whether or not there is a video recording. So the extra file in the test is due to something like this:
1. The 1st stop -> the 1st file
2. The 2nd stop -> the 2nd file
3. afterSession -> cleanupNode -> the 3rd file with the same testInformation as the 2nd

This only happen if your test use the cookie to disable recording right before the afterSession. If you don’t do that and let the last test finishes on its own(which is a proper thing to do since there is no point in using the cookie right before the session ends), then you would get one item as expected. For the IT, you would get 2 files if I don’t do the 2nd recording stop.

On a similar note, I notice that you also get a duplicate item (different name with “_{num}”) in the dashboard if a test is timed out and then the session ends due to this:
1. shutdownIfStale is triggered and you get 1 item in the dashboard
2. afterSession is triggered and you get a duplicate item.

I fixed this by only allowing cleanupNode call if the isCleaninup is false.

Regarding Travis, I see that the dockerCompose one also failed, probably due to the parallel nature of the test. I modified the test a little bit and tried to do a different assertion. I’d like to get the dockerCompose one to pass first to narrow down the issue. Can you please try this one?

ccharnkij:
@diemol Looks like docker-compose test went well. It had issue removing the video file during 'clean' after test. The cloud ones still have issue. For tests like 'sauceLabs', is it possible that the test was ran thru SauceLabs?

diemol:
Sadly you cannot commit directly to the pr-1085 branch because you have no permissions on this project, and even worse, I do not have admin permissions anymore since I am not part of Zalando anymore since a while ago.
Perhaps you can just, as suggested, move the test to the `minikube` group and keep working on this same branch.

ccharnkij:
@diemol I added the minikube group. Please try again.

diemol:
@ccharnkij no need to wait for me, you can now already see if the test is passing or not, it always runs when it is part of the `minikube` group.
E.g., it is now failing, see [here](https://travis-ci.org/zalando/zalenium/jobs/645269283#L5268)
Just iterate on it.

ccharnkij:
@diemol Finally!!!

diemol:
@ccharnkij could you please update your fork and get the latest changes from master? After that I can have a look between today and tomorrow. (Sorry, I am swamped with work)

ccharnkij:
@diemol No worries. I rebased and squashed all my commits together as well.

diemol:
Hi @ccharnkij,

I was trying the PR locally again, and I believe the result is not the expected one. I can see how the test passes, but I think the test can be improved. We should have a test where the video is stopped and started a few times (which is the end goal of the feature).

So I went and I modified the test, check this [commit](https://github.com/ccharnkij/zalenium/commit/4acafc1df8ba0f933ad3f6174982010e9bfb3dba), and sadly I was not able to see the feature working properly. Could you please run the code locally and have a look?  

I am sorry I have not had time lately to look into this, but I have reserved a few hours this week to do this.

ccharnkij:
@diemol I see what you are trying to do. I got it working. Let me know how this one goes.

diemol:
Just wanted to add that I added a few TestNG group annotations to execute the tests in the right jobs, which made the tests stable when the cloud providers integrations are being tested as well.

PR #1104, opened 2019-12-05
feature(k8s): enable only adding some videos to persisted volume

alfonso-presa:
### Description
This enables changing the mount path of the videos volume in eks.

### Motivation and Context
For auditing purposes (due to legal requirement) we need to save evidences (videos) of tests executed prior to deployment to some environments for a large amount of time (>2 years). We execute plenty of tests that we don't need the videos to be persisted. With this change, by adding the `zal:build` capability we are able to make those test evidences end in an specific folder inside `/home/seluser/video` called `evidences`. We can take advantage of the current video PVC to provide an NFS to S3 FileGateway volume so that the videos of our interest end up in an S3 bucket. We don't mind loosing the videos in the dashboard upon servers restarts (although if you want we could provide another optional PVC for the purpose of this PR, we did not in shake of getting the minimal code impact to the project). 

### How Has This Been Tested?
We created and NFS to S3 File Gateway Claim and added it to our helm by using the values file and the properties for that purpose. The dashboard keep showing the videos and logs properly.

### Types of changes

### Checklist:
- [X] My change requires a change to the documentation.
- [X] I have updated the documentation accordingly.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -391,6 +391,8 @@ COPY zalenium.sh \
  /home/seluser/
 COPY scm-source.json /
 
+RUN mkdir /home/seluser/videos
+
 #-----------------#
 # Fix perms again #
 #-----------------#
--- a/charts/zalenium/templates/_pod-template.yaml
+++ b/charts/zalenium/templates/_pod-template.yaml
@@ -164,7 +164,7 @@ spec:
         {{- toYaml .Values.hub.resources | nindent 8 }}
       volumeMounts:
         - name: {{ template "zalenium.fullname" . }}-videos
-          mountPath: /home/seluser/videos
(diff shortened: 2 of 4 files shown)


docker/Dockerfile
```diff
@@ -391,6 +391,8 @@ COPY zalenium.sh \
  /home/seluser/
 COPY scm-source.json /
 
+RUN mkdir /home/seluser/videos
```
diemol: Why do we need to create this directory when the image is built? This should be done implicitly when the volume is mounted.

alfonso-presa: Because with this changes you might use a subfolder of videos for persistance.... So the folder is needed for the process not to fail.

PR #1169, opened 2020-03-26
Update KubernetesContainerClient.java readiness probe

tomassatka:
Could you please remove proxy from readinessProbe? I dont see a need for it in this probe...
Please correct me if i am wrong and the check is required and has its purpose.

quickly.**

### Description

Removing proxy from readiness probe

### Motivation and Context

If proxy is set over env configuraton as zalenium_http_proxy then readyness probe fails. After 2 minutes pods are killed and restarted. 

### How Has This Been Tested?

Only locally

### Types of changes

- [ x] Bug fix (non-breaking change which fixes an issue)

### Checklist:

- [ ] My code follows the code style of this project.
- [ ] I have read the **CONTRIBUTING** document.

### DIFF
--- a/src/main/java/de/zalando/ep/zalenium/container/kubernetes/KubernetesContainerClient.java
+++ b/src/main/java/de/zalando/ep/zalenium/container/kubernetes/KubernetesContainerClient.java
@@ -679,7 +679,7 @@ public class KubernetesContainerClient implements ContainerClient {
                         // so then we can initiate a registration.
                         .withNewReadinessProbe()
                             .withNewExec()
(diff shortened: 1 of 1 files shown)


pearj:
That is simply disabling the proxy during the health check, because we don't want a proxy to get inbetween zalenium and the nodes.

pearj:
Also, this project has had it's [last release](https://github.com/zalando/zalenium/releases/tag/3.141.59z), there will be no more merged pr's, so you'll have to build zalenium yourself if you want this change.

pearj:
Personally, I have found proxy access to be much more reliable if you don't use zalenium_http_proxy, but rather set http proxy via selenium capabilities.

PR #1175, opened 2020-04-07
api version updated

ckavili:
quickly.**

### Description

apiVersion for route object is updated as route.openshift.io/v1
### Motivation and Context

The API version in the template (v1) does not match the expected API version (route.openshift.io/v1) when trying to install the chart with ArgoCD. So it leads to ArgoCD to not create the route.
### How Has This Been Tested?

Route is deployed to OCP v4 cluster through ArgoCD and helm cli when hub.openshift.route.enabled set true.

### Types of changes

### Checklist:

### DIFF
--- a/charts/zalenium/templates/route.yaml
+++ b/charts/zalenium/templates/route.yaml
@@ -1,5 +1,5 @@
 {{- if .Values.hub.openshift.route.enabled -}}
-apiVersion: v1
+apiVersion: route.openshift.io/v1
 kind: Route
 metadata:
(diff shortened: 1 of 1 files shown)


diemol:
Wouldn't this change break things for all other people who are using the chart?

ckavili:
Nope, it wouldn't. `v1` is the core group and because API group is not specified,  Kind matches with`route.openshift.io/v1` since it’s in the core API groups. It's always better to use the specific one where possible.

pearj:
That sounds reasonable to me

PR #1209, opened 2020-10-21
Unifies SECURITY.md to point to the Zalando security form

bocytko:
Unifies SECURITY.md to point to the Zalando security form.


diemol:
👋  @bocytko 
Could we please archive this project?

